PR #6, opened 2023-07-17
feat: establish SB connection on 1st message delivery

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-171

### What's the issue?
Previously, we established a connection to the Sendbird SDK upon mounting the <SendbirdProvider />. 
However, this approach resulted in charging users even when they hadn't sent any messages in the chat room.

### How did I handle it?
To address this issue, I made users initially see a fake channel (using the `<StartingPage />` component) until they send their first message. Once the connection is successfully established, we will switch the channel to the real one by mounting `<CustomChannelComponent />`.

https://github.com/sendbird/chat-ai-widget/assets/10060731/3f1cdd47-a975-4e6e-8869-336c04f61cf4


luke-cha (approved):
LGTM

PR #12, opened 2023-07-21
feat: apply appStateToggleEnabled init param

AhyoungRyu:
Even after https://github.com/sendbird/chat-ai-widget/pull/11 merged, the channel is still refreshed due to the sdk reconnection. It's because the sdk's default behavior https://github.com/sendbird/chat-js/blob/develop-v4/src/comm/appStateChangeDetector.ts#L26. So I had to add a new option; `sdkInitParams` to UIKit https://github.com/sendbird/sendbird-uikit-react/pull/692 so we can pass the sdk init option param.

By passing [`appStateToggleEnabled: false`](https://sendbird.com/docs/chat/v4/javascript/ref/interfaces/_sendbird_chat.SendbirdChatParams.html#appStateToggleEnabled) when `sdk.init()` is called, we'll be able to prevent this unnecessary channel refresh.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.1",
       "dependencies": {
         "@sendbird/chat": "^4.9.3",
-        "@sendbird/uikit-react": "^3.6.4",
+        "@sendbird/uikit-react": "3.6.5",
         "dompurify": "^3.0.4",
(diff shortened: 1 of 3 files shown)


luke-cha (approved):
LGTM!

PR #13, opened 2023-07-21
Fix build script to exclude when build and set text align

luke-cha:
image

### What's the issue?
[Slack](https://sendbird.slack.com/archives/C0593MY7KMG/p1689924369178099)
In version 1.0.1, if built while .env file exists, there is an issue where the default `applicationId` and `botId` values get set

### How did I solve it?
Added logic to change `.env` to `.env_temp` pre build, and then revert it back post build. Additionally, fixed the issue where the chat message sender's text_align was set to center.


AhyoungRyu (approved):
Thanks Luke!

PR #20, opened 2023-07-27
Fix pending after first message

luke-cha:
### Details
Add detailed waiting logic for the issue that Bot's message is pending after sending the first message

### How to
Add logic to wait until the last message to the group channel is not null (up to 3 seconds) because it determines that an error occurs when the channel loads immediately after sending a message to the group channel

### DIFF
--- a/src/hooks/useCreateGroupChannel.ts
+++ b/src/hooks/useCreateGroupChannel.ts
@@ -12,6 +12,19 @@ import { useConstantState } from '../context/ConstantContext';
 import { useSbConnectionState } from '../context/SBConnectionContext';
 import { delay } from '../utils';
 
+async function waitForLastMessage(
+  channel: GroupChannel,
+  maxRetries = 30,
+  retryInterval = 100
+) {
+  let count = 0;
+  while (channel.lastMessage == null && count < maxRetries) {
+    await delay(retryInterval);
+    count++;
+  }
+  await delay(500);
+}
+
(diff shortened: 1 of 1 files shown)


src/hooks/useCreateGroupChannel.ts
```diff
@@ -49,8 +49,12 @@ export function useCreateGroupChannel(
           message: firstMessage,
         });
       }
-      // let's wait for a second to make sure channel is created & ready
-      await delay(1000);
+      let count = 0;
+      while (groupChannel.lastMessage == null && count < 30) {
+        await delay(100);
+        count += 1;
+      }
```
AhyoungRyu: nit;

```
async function waitForLastMessage(groupChannel, maxRetries = 30, retryInterval = 100) {
  let count = 0;
  while (groupChannel.lastMessage == null && count < maxRetries) {
    await delay(retryInterval);
    count++;
  }
  await delay(500);
}
```
How about having this logic separately and call it in here? 
```
// usage 
await waitForLastMessage(groupChannel);
```

luke-cha: @AhyoungRyu Your suggestion is also good. Could you update it?

AhyoungRyu: Yeah sure will do.

AhyoungRyu (approved):
Thank you, Luke!

PR #21, opened 2023-07-27
feat: add customRefreshComponent prop

AhyoungRyu:
### What's the issue?
1. Currently, the refresh icon and the expand + close icons are placed separately, making it difficult to control them neatly. As a result, we use `<EmptyContainer />` to avoid overlapping the refresh icon with the close icon.
2. There's currently no way to handle the refresh icon from outside of the ChatWidget component.

### How I solved it:
1. Removed the EmptyContainer that was previously positioned next to the refresh icon.
2. Introduced a new prop: `customRefreshComponent`, which allows the chat-ai-widget user to control the refresh icon by providing their own custom icon component.

```typescript
export interface CustomRefreshComponent {
  icon: React.FC;
  width: string;
  height: string;
  onClick?: () => void;
  style?: React.CSSProperties;
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,7 @@ const App = () => {
       chatBottomContent={customConstants.chatBottomContent}
       messageBottomContent={customConstants.messageBottomContent}
       replacementTextList={customConstants.replacementTextList}
+      customRefreshComponent={customConstants.customRefreshComponent}
     />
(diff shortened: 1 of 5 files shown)


luke-cha (approved):
LGTM

PR #26, opened 2023-08-01
feat: apply lazy loading for CodeBlock

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-259 

### What's the issue?
One of our users reported an issue where the chat AI widget takes some time to load the `react-code-blocks` library during initialization. 
This delay is due to the relatively large bundle size of `react-code-blocks`, and it appears that the library isn't loaded until it's required, i.e. when a message contains a code block format

### How did I solve it?
Addressed a lazy loading for the `<CodeBlock />` component by separating it from `<ParsedBotMessageBody />`.
Now, the `react-code-blocks` library is loaded only when necessary, resulting in a more efficient and faster loading process for the chat AI widget.

It also creates a separate bundle like this.
Screenshot 2023-08-01 at 3 56 27 PM

### DIFF
--- a/src/components/CodeBlock.tsx
+++ b/src/components/CodeBlock.tsx
@@ -1,0 +1,16 @@
+// Ref: https://github.com/rajinwonderland/react-code-blocks#-demo
+import { CopyBlock as Component, irBlack } from 'react-code-blocks';
+
+import { Token } from '../utils';
+
+export function CodeBlock({ token }: { token: Token }) {
+  return (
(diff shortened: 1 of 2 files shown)


AhyoungRyu:
Screenshot 2023-08-01 at 3 47 06 PM

Confirmed that the code block itself is working fine as before.

luke-cha (approved):
Good!

PR #30, opened 2023-08-08
feat: mobile support

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-289

### DIFF
--- a/src/components/CustomChannelHeader.tsx
+++ b/src/components/CustomChannelHeader.tsx
@@ -4,6 +4,7 @@ import styled from 'styled-components';
 import { useConstantState } from '../context/ConstantContext';
 import { useSbConnectionState } from '../context/SBConnectionContext';
 import channelHeaderImage from '../icons/bot-message-image.png';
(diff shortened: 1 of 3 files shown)


src/components/CustomChannelHeader.tsx
```diff
@@ -88,7 +89,7 @@ export default function CustomChannelHeader(props: Props) {
           }}
         />
         <Title>{channel.name}</Title>
-        {(betaMark || customBetaMarkText) && (
+        {!isMobile && (betaMark || customBetaMarkText) && (
```
AhyoungRyu, line 92: **Why we have this condition?**
-> I wanted to tweak this overlapped elements style when the header's width is too narrow.

Screenshot 2023-08-08 at 4 29 20 PM

sravan-s: cc @liamcho  any idea?

sravan-s (approved):

PR #35, opened 2023-08-11
fix: add firstMessage to createChannel hook dependency list

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-304 

### What's the issue?
When the refresh icon is clicked, the [two functions below are executed](https://github.com/sendbird/chat-ai-widget/blob/develop/src/components/CustomChannelHeader.tsx#L74):  

```
  setFirstMessage(null);
  createGroupChannel(); 
```

The `createGroupChannel` custom hook uses the `firstMessage` inside the effect function. 
However, the `firstMessage` state has not been included in the effect's dependency array, causing the initialized state not to be reflected correctly. 

### How did I fix it? 
I simply added the missing `firstMessage` to the dependency array.


chohongm (approved):
lg2m!

PR #37, opened 2023-08-14
feat: add quick reply UI

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-314

https://github.com/sendbird/chat-ai-widget/assets/10060731/4231957f-0392-45be-aeba-544a96459f0e

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - develop
-  pull_request:
 
 jobs:
   build-and-deploy:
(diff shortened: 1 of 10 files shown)


.github/workflows/deploy.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - develop
-  pull_request:
```
AhyoungRyu, line 7: I removed it because I thought we'd better publish the github pages only when `develop` branch is updated.

sravan-s (approved):

PR #43, opened 2023-08-24
feat: attatch quick reply UI to bottom of bot message

AhyoungRyu:
Address https://sendbird.atlassian.net/browse/AC-356

**AS-IS**
Screenshot 2023-08-24 at 6 12 23 PM
The quick reply component was used to be at the top of the message input box but it makes the scroll behavior flaky. 

**TO-BE**
Screenshot 2023-08-24 at 5 24 51 PM
So I just moved the quick reply component to the bottom of bot message.


luke-cha (approved):
LGTM!

PR #44, opened 2023-08-25
fix: incorrect quick reply format

AhyoungRyu:
Fixes https://sendbird.atlassian.net/browse/AC-364

Should be consistent when user pass the firstMessageData by modifying the source code or changing the setting via Sendbird dashbord. 

- When the first message & quick reply data are set via dashboard
Screenshot 2023-08-25 at 2 19 27 PM

- When the values are set via source code
Screenshot 2023-08-25 at 2 22 17 PM

Both cases should work as expected

### DIFF
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -2,9 +2,12 @@ import React from 'react';
 import ReactDOM from 'react-dom/client';
 
 import App from './App';
+import { DEFAULT_CONSTANT } from './const';
 
 ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
   <React.StrictMode>
-    <App />
+    <App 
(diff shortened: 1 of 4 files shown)


src/main.tsx
```diff
@@ -2,9 +2,12 @@ import React from 'react';
 import ReactDOM from 'react-dom/client';
 
 import App from './App';
+import { DEFAULT_CONSTANT } from './const';
 
 ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
   <React.StrictMode>
-    <App />
+    <App 
+      firstMessageData={DEFAULT_CONSTANT.firstMessageData}
```
AhyoungRyu, line 10: More init prop values should be set from here not from the ConstantContext.

luke-cha (approved):

PR #47, opened 2023-08-30
feat: introduce ai widget script loading for non-react users 

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-376 

### Why we're doing this?
We are making these changes to assist users who do not use React or any other JavaScript framework or library. Our aim is to enable them to utilize this widget library without any additional dependencies, simply by loading the script from their website (HTML file).

Here's a working example: 
Screenshot 2023-08-30 at 5 15 28 PM

Additionally, I have included `js-example.html` as a quick reference example.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -33,9 +33,15 @@ export default defineConfig({
       external: [
         'react',
         'react-dom',
-        'styled-components',
         'react-code-blocks',
       ],
+      output: {
+        globals: {
+          react: 'React',
+          'react-dom': 'ReactDOM',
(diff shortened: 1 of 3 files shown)


vite.config.ts
```diff
@@ -33,9 +33,15 @@ export default defineConfig({
       external: [
         'react',
         'react-dom',
-        'styled-components',
```
AhyoungRyu, line 36: It seemed to be okay to remove this line but will double check when I create a RC release.

luke-cha (approved):
Good~

PR #49, opened 2023-09-04
fix: set scroll block option to 'end' from 'nearest'

AhyoungRyu:
Just realized that the scroll block option should be 'end' to make sure the scroll downs to the end on streaming. 
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#parameters


luke-cha (approved):
Added an intro GIF change for Readme.

PR #54, opened 2023-09-08
feat: Add boolean parameter to control emoji attachment

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-461

>After adding https://sendbird.atlassian.net/browse/AC-454,  there are probably some users who don't want emojis attached to all bot messages. Therefore, we should add a boolean parameter to allow them to control this feature.


AhyoungRyu:
The CircleCI failure is related to #51 I guess not this PR.

bang9 (approved):

PR #50, opened 2023-09-06
feat: Merge sender profile when message is sent in short span of time

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-438

**AS-IS** 
Display the profile image & bot nickname for every single message 
Screenshot 2023-09-06 at 6 01 45 PM

**TO-BE**
Display only the first message's bot nickname & last one's profile image within the certain span of time(3min in our case)
Screenshot 2023-09-06 at 5 57 55 PM

PR #55, opened 2023-09-11
fix: make profile img invisible only when consecutive msg senders are same

AhyoungRyu:
### What's the issue?
After #50 merged, the bot's profile image is invisible even though the consecutive messages' senders are different. 
Screenshot 2023-09-11 at 1 13 58 PM

The 1st bot message's profile image & 2nd bot message's nickname in the screenshot are invisible which is not the expected behavior.

### With this patch 
The 1st bot message's profile image & 2nd bot message's nickname in the screenshot should be visible. 
Screenshot 2023-09-11 at 1 14 29 PM

### DIFF
--- a/src/utils/messages.ts
+++ b/src/utils/messages.ts
@@ -13,7 +13,7 @@ export function groupMessagesByShortSpanTime(
 ): EveryMessage[] {
   // Create an object to group messages based on their creation time
   const groupedMessagesByCreatedAt = messages.reduce((groups, message, idx) => {
-    const { createdAt, sender } = message;
+    const { createdAt, sender, messageType } = message;
     // Get the key of the previous group
     const prevKey = Object.keys(groups)[Object.keys(groups).length - 1];
 
@@ -21,6 +21,7 @@ export function groupMessagesByShortSpanTime(
     if (
       prevKey &&
       createdAt - Number(prevKey) <= TIME_SPAN &&
+      messageType !== 'admin' &&
       sender?.userId === messages[idx - 1]?.sender.userId
     ) {
       // Add the message to the existing group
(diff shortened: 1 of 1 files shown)


src/utils/messages.ts
```diff
@@ -12,13 +12,17 @@ export function groupMessagesByShortSpanTime(
...
     // Get the key of the previous group
     const prevKey = Object.keys(groups)[Object.keys(groups).length - 1];
 
     // Check if the time difference between the current message and the previous one is within 3 minutes
-    if (prevKey && message.createdAt - Number(prevKey) <= TIME_SPAN) {
+    if (
+      prevKey &&
+      createdAt - Number(prevKey) <= TIME_SPAN &&
+      sender?.userId === messages[idx - 1]?.sender.userId
+    ) {
```
bang9, line 26: Can admin messages be sent through the AI widget as well?
In RN UIKit grouping, there is filtering for admin messages(since admin messages don't have a sender), so I'm just checking to see if it's possible.

bang9 (approved):
lgtm!

src/utils/messages.ts
bang9: For example

```
// GOOD
(profile)Bot
                         Admin
(profile)Bot
                                        User(-------)
                                        User(profile)
(-------)Bot
(profile)Bot
```

```
// BAD
(-------)Bot
                         Admin
(profile)Bot
                                        User(-------)
                                        User(profile)
(-------)Bot
(profile)Bot
```

luke-cha (approved):
LGTM!

src/utils/messages.ts
AhyoungRyu: Oh thanks Airen! https://github.com/sendbird/chat-ai-widget/pull/55/commits/7fab34ba1214ecad6d05206b2b1522f89d5aaa62 just updated the condition to filter out the admin message.

bang9: ++ if that's the case, we should add optional chaining to sender

```diff
- messages[idx - 1]?.sender.userId
+ messages[idx - 1]?.sender?.userId
```

PR #56, opened 2023-09-12
feat: modify emoji display condition

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-476

**AS-IS**

**TO-BE**

1. The emoji shouldn’t be displayed on the bot welcome messages 
2. The emoji should be displayed only for the last one if the bot is sending multiple consecutive messages

### DIFF
--- a/src/components/CustomMessage.tsx
+++ b/src/components/CustomMessage.tsx
@@ -1,7 +1,6 @@
 import { User } from '@sendbird/chat';
 import { UserMessage } from '@sendbird/chat/message';
 import { useChannelContext } from '@sendbird/uikit-react/Channel/context';
-import { useMemo } from 'react';
 // eslint-disable-next-line import/no-unresolved
 import { EveryMessage } from 'SendbirdUIKitGlobal';
 
@@ -21,7 +20,6 @@ import {
   replaceUrl,
   Token,
 } from '../utils';
-import { getBotWelcomeMessages } from '../utils/messages';
 
 type Props = {
   message: EveryMessage;
@@ -30,6 +28,7 @@ type Props = {
   lastMessageRef: React.RefObject<HTMLDivElement>;
   chainTop?: boolean;
   chainBottom?: boolean;
+  isBotWelcomeMessage: boolean;
 };
 
 export default function CustomMessage(props: Props) {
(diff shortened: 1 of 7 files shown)


src/components/ParsedBotMessageBody.tsx
```diff
@@ -81,7 +81,7 @@ export default function ParsedBotMessageBody(props: Props) {
             </BlockContainer>
           );
         })}
-        {sources.length > 0 ? (
+        {sources.length > 0 && enableSourceMessage ? (
```
AhyoungRyu, line 84: Just found that this condition needs to be here not on the above to make it work correctly.

src/const.ts
```diff
@@ -50,15 +50,15 @@ export const DEFAULT_CONSTANT: Constant = {
     ],
   },
   firstMessageData: [
```
AhyoungRyu, line 52: Disabled this line cause I don't think it's really helpful 😅

src/components/CustomMessage.tsx
```diff
@@ -45,6 +47,16 @@ export default function CustomMessage(props: Props) {
...
+  const isBotWelcomeMessage = useMemo(() => {
+    const botWelcomeMessages = getBotWelcomeMessages(
+      allMessages,
+      botUser.userId
+    );
+    return !!botWelcomeMessages.find(
+      (welcomeMessage) => welcomeMessage.messageId === message.messageId
+    );
+  }, [allMessages.length]);
+
```
bang9: ~~How about this instead of searching for all the messages?~~

```suggestion
  const isBotWelcomeMessage = message.sender?.userId === botUser.userId // isBotMessage(message, botUser);
```

AhyoungRyu: Oh I was trying to get only the welcoming messages not the all bot messages.
Screenshot 2023-09-12 at 5 21 22 PM

So these are the welcome messages.

bang9: Ahh,, I understand why it's a "welcome message" ㅋㅋㅋ

How about lifting `botWelcomeMessages` up to a higher level rather than recalculating it in each message component?

bang9 (approved):

src/components/CustomMessage.tsx
AhyoungRyu: Yeah sounds good. https://github.com/sendbird/chat-ai-widget/pull/56/commits/ca4fa47fa9f78510df395d9b2301c7fffdfbb67a Just addressed it :)

chohongm (commented):
LG2M

PR #63, opened 2023-10-11
[AC-558] Support  3>= users conversation in the chatbot

AhyoungRyu:
Addresses https://sendbird.atlassian.net/browse/AC-558 

We've only allowed bot <-> single user conversation in the chatbot but we want to allow 3 >= users conversation too. 
Screenshot 2023-10-10 at 4 52 05 PM


.env
```diff
@@ -1,12 +1,4 @@
 # Vite prefix is required for Vite to load the env variables
 # Plz modify below two env variables on your needs
-#VITE_CHAT_WIDGET_APP_ID=CEA3E63E-BBA0-4188-A181-0985DE7BB350
```
AhyoungRyu, line 3: This shouldn't be updated so I put it back 😅

bang9 (approved):

PR #66, opened 2023-11-09
fix: welcome message URL parsing

AhyoungRyu:
Fixes https://sendbird.atlassian.net/browse/AC-753

Just removed unnecessary condition for the welcome message. It doesn't need to be handled separately than the other normal messages.
Confirmed that a URL link in the first message is being correctly parsed. 
Screenshot 2023-11-09 at 3 31 33 PM


luke-cha (approved):
LGTM

PR #72, opened 2023-12-11
feat: exclude pull request from github.io deploy option

AhyoungRyu:
https://sendbird.github.io/chat-ai-widget/ shouldn't be updated on every pull requests but only accept develop branch.


luke-cha (approved):
Good!

PR #73, opened 2024-01-08
feat: export type def

AhyoungRyu:
> Note: Need to publish a new version to take advantage of this change. 

**Before**
Screenshot 2024-01-08 at 9 34 00 AM

**After**
Tried my local build.
Screenshot 2024-01-08 at 9 28 06 AM


luke-cha (approved):
LGTM!

PR #77, opened 2024-02-07
Fix autoOpen & betaMark behavior

AhyoungRyu:
Just fixed `autoOpen` & `betaMark` props are not correctly working as expected due to the incorrect condition.


luke-cha (approved):
LGTM

PR #82, opened 2024-02-14
Init self-service directory

AhyoungRyu:
Address https://sendbird.atlassian.net/browse/AC-1242

### DIFF
--- a/packages/self-service/embed-example.html
+++ b/packages/self-service/embed-example.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Vite + React + TS</title>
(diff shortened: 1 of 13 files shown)


packages/self-service/embed-example.html
```diff
@@ -0,0 +1,26 @@
...
+    <title>Vite + React + TS</title>
+    <link rel="preload" href="https://fonts.cdnfonts.com/css/gellix" rel="stylesheet">
+    <link rel="preload" href="https://fonts.cdnfonts.com/css/sf-pro-display" rel="stylesheet">
+  </head>
+  <body>
+    <div id="aichatbot"></div>
+    <link href="/dist/index.css" rel="stylesheet" />
+    <script>
+      window.chatbotConfig = {
+        appId: "125B2432-BB0A-4650-95E2-80810404D526",
```
chohongm, line 16: is it ok to expose this

chohongm (approved):
LGTM

packages/self-service/embed-example.html
AhyoungRyu: Yeah it's just one of demo app & bot ids.

PR #95, opened 2024-02-27
feat: replace Channel module \w GroupChannel

AhyoungRyu:
Some of UIKit related issues have been solved in the new GroupChannel module(which has been introduced in react-uikit v3.12.0). So I replaced Channel -> GroupChannel to take an advantage of it. 

This will fix this issue too https://sendbird.atlassian.net/browse/EI-906


luke-cha (approved):
Good!

PR #99, opened 2024-02-28
feat: update url-webdemo configurations

AhyoungRyu:
To keep consistency with the old widget style in sendbird.com/docs, needed to modify some configs such as 
- enableMention: true -> false
- enableEmojiFeedback:  true -> false 
- enableSourceMessage: false -> true


luke-cha (approved):
LGTM

PR #101, opened 2024-02-29
[EL-909] Improve initial rendering

AhyoungRyu:
- [x] used useQuery in `useGroupChannel` not to call create group channel API more than it's necessary. And dropped `useCreateGroupChannel` which is using useEffect internally
- [x] set `staleTime` in QueryClient to prevent duplicated channel style API in the beginning 
- [x] removed hashedKey(this is from url-webdemo. So don't need to keep in our core logic but it's okay to manage just within the url-webdemo package) related logic and manage the hashedKey like botId  since they're exactly the same
- [x] removed useGetBotUser hook which calls user list query API internally. We don't need this but we can just use the useGroupChannel hook since we can get the bot user information in the group members' info. 
- [x] Confirm there's no side effect on original behavior e.g. initial channel creation / message sending / bot welcome message is sent when the new channel is created...

#### Before
https://github.com/sendbird/chat-ai-widget/assets/10060731/90a4ab34-6cd7-4818-992f-a4f65f3691e1

1.38s -> 3.92s = about 2.5s

#### After

https://github.com/sendbird/chat-ai-widget/assets/10060731/ec1f13a6-2a8a-407a-8ea9-37535daa0326

1.4 -> 2.7s = about 1.3s 

> NOTE: To test this, I used an appId created in Seoul region


src/components/CustomChannel.tsx
```diff
@@ -1,40 +1,27 @@
...
 
 export default function CustomChannel() {
-  const { botId, instantConnect } = useConstantState();
-  const store = useSendbirdStateContext();
-  const sb: SendbirdGroupChat = store.stores.sdkStore.sdk as SendbirdGroupChat;
-
+  const { botId } = useConstantState();
   assert(botId !== null, 'botId must be provided');
 
-  const botUser: User = useGetBotUser(sb.currentUser, botId) as User;
```
AhyoungRyu, line 20: I removed this `useGetBotUser` hook which calls user list query API internally to prevent an additional API call in the beginning. 

Instead, we can just use the useGroupChannel query hook since we can get the bot user information in the group members' info like [this way](https://github.com/sendbird/chat-ai-widget/pull/101/files#diff-3aa01ac34a251f9259d802040a2d9ece492d643beb8eadf7ed046351b29e0630R34).

src/components/CustomChannelHeader.tsx
```diff
@@ -63,13 +64,13 @@ const RenewButtonContainer = styled.div`
 type Props = {
   channel: GroupChannel;
   botUser: User;
-  createGroupChannel: () => void;
 };
 
 export default function CustomChannelHeader(props: Props) {
-  const { channel, createGroupChannel, botUser } = props;
```
AhyoungRyu, line 70: Removed prop drilling but get this from useGroupChannel query hook

src/components/ChatAiWidget.tsx
```diff
@@ -162,7 +162,7 @@ const Component = (props: Props) => {
 
   useEffect(() => {
     if (props.autoOpen || channelStyle.autoOpen) {
-      timer.current = setTimeout(() => setIsOpen(() => true), 1000);
```
AhyoungRyu, line 165: Reduced the automatic chat window loading time 1000ms -> 200ms

luke-cha (approved):
Good job

src/components/ChatAiWidget.tsx
```diff
@@ -161,8 +161,8 @@ const Component = (props: Props) => {
   };
 
   useEffect(() => {
-    if (props.autoOpen || channelStyle.autoOpen) {
-      timer.current = setTimeout(() => setIsOpen(() => true), 1000);
+    if (!isMobile && (props.autoOpen || channelStyle.autoOpen)) {
```
luke-cha, line 164: I made changes to the code because there were instances where isOpen became true after 0.2 seconds, regardless of whether it was on mobile or not.

AhyoungRyu: Thanks!

PR #110, opened 2024-03-05
fix: typing indicator image format change

luke-cha:
Description: The typing indicator image was a GIF, causing file size issues during the build process. It has been changed to WebP.


bang9 (approved):
lgtm!

PR #112, opened 2024-03-05
chore: add self-service script publish guide

AhyoungRyu:
Added a self-service script deployment guide just for quick reference. More detailed instructions will be added to Jira Confluence in the near future!

### DIFF
--- a/packages/self-service/embed-example.html
+++ b/packages/self-service/embed-example.html
@@ -19,9 +19,9 @@
         j = d.createElement(s);
         j.defer = true;
         j.type = 'module';
-        j.src = 'https://aichatbot.sendbird.com/index.js';
+        j.src = './dist/index.js';
         f.parentNode.insertBefore(j, f);
-      }(window, document, 'script', 'A3E33A85-E4A7-46A4-99B1-EF6833E7DE3A', 'marketing_bot_3');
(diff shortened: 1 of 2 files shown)


packages/self-service/embed-example.html
```diff
@@ -19,9 +19,9 @@
         j = d.createElement(s);
         j.defer = true;
         j.type = 'module';
-        j.src = 'https://aichatbot.sendbird.com/index.js';
+        j.src = './dist/index.js';
         f.parentNode.insertBefore(j, f);
-      }(window, document, 'script', 'A3E33A85-E4A7-46A4-99B1-EF6833E7DE3A', 'marketing_bot_3');
```
AhyoungRyu, line 24: Will it be better if I just keep this? Otherwise we might need to look for  appid / botid everytime from somewhere else 😅

Or I can use the Khan academy one in the .env.

bang9: For testing purposes, it looks good to keep this!
Additionally, to prevent issues in recent deployments, it would be good to have apps and files per plan (starter/pro). For example, embed-example-pro.html and embed-example-starter.html.

bang9 (approved):
lgtm~

packages/self-service/embed-example.html
AhyoungRyu: Yeah sounds good. I'll prepare them too in a separate PR :)

PR #113, opened 2024-03-06
chore: add prebuild script for npm build

bang9:


AhyoungRyu (approved):
Thank you for doing this! 
Probably we also need to replace `npm run build` -> `npm run build:npm` in `release-guide.md` file.

PR #116, opened 2024-03-11
[CLNP-2532] fix: html xss

bang9:
- replace `dangerouslySetInnerHTML` with `HTML tag` to `RegexText` component
- support markdown URL format(`[text](url)`)
- support markdown bold format(`**bold**`)

ticket: [CLNP-2532]

[CLNP-2532]: https://sendbird.atlassian.net/browse/CLNP-2532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,18 +31,19 @@
   "dependencies": {
     "@sendbird/chat": "^4.10.1",
     "@sendbird/uikit-react": "^3.13.1",
+    "@tanstack/react-query": "^5.17.19",
     "dompurify": "^3.0.4",
     "polished": "^2.3.1",
     "react-code-blocks": "^0.1.0",
     "react-popper-tooltip": "^4.4.2",
-    "styled-components": "^5.3.11",
-    "@tanstack/react-query": "^5.17.19"
+    "styled-components": "^5.3.11"
   },
   "devDependencies": {
+    "@tanstack/eslint-plugin-query": "^5.17.22",
+    "@types/dompurify": "^3.0.5",
     "@types/react": "^18.0.37",
     "@types/react-dom": "^18.0.11",
     "@types/styled-components": "^5.1.26",
-    "@tanstack/eslint-plugin-query": "^5.17.22",
     "@typescript-eslint/eslint-plugin": "^5.60.1",
     "@typescript-eslint/parser": "^5.60.1",
     "@vitejs/plugin-react": "^4.0.0",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -31,18 +31,19 @@
...
     "polished": "^2.3.1",
     "react-code-blocks": "^0.1.0",
     "react-popper-tooltip": "^4.4.2",
-    "styled-components": "^5.3.11",
-    "@tanstack/react-query": "^5.17.19"
+    "styled-components": "^5.3.11"
   },
   "devDependencies": {
+    "@tanstack/eslint-plugin-query": "^5.17.22",
+    "@types/dompurify": "^3.0.5",
```
AhyoungRyu, line 43: 👍

AhyoungRyu (approved):
I tested and it works well. Thanks!

src/components/ParsedBotMessageBody.tsx
```diff
@@ -90,3 +123,45 @@ export default function ParsedBotMessageBody(props: Props) {
...
+  children: string;
+  patterns: RegexTextPattern[];
+}) => {
+  if (patterns.length === 0 || typeof children !== 'string') {
+    return <>{children}</>;
+  }
+
+  const convertedNodes: Array<string | ReactNode> = [children];
+  patterns.forEach(({ regex, replacer }) => {
+    const node = convertedNodes.concat();
```
chohongm, line 159: what is point of concat without arg?

src/components/ParsedBotMessageBody.tsx
```diff
@@ -90,3 +123,45 @@ export default function ParsedBotMessageBody(props: Props) {
...
+    let offset = 0;
+    node.forEach((text, index) => {
+      if (typeof text === 'string' && text) {
+        const children = replaceWithRegex(text, regex, replacer);
+
+        if (children.length > 1) {
+          convertedNodes.splice(index + offset, 1, ...children);
```
chohongm, line 166: children is a string right? why we need spread operator here?

src/components/ParsedBotMessageBody.tsx
bang9: It just copies the array
it's the same as `const node = [...convertedNodes];`

src/components/ParsedBotMessageBody.tsx
bang9: children is a `Array<string | ReactNode>` type

PR #122, opened 2024-03-19
[AC-1660] Drop react-code-block & add new CodeBlock component

AhyoungRyu:
In terms of reducing redundant pkg dependency, dropping react-code-block library but adding a new CodeBlock component  instead. 

record
^ Here's how it works.

### DIFF
--- a/src/components/CodeBlock.tsx
+++ b/src/components/CodeBlock.tsx
@@ -1,16 +1,82 @@
-// Ref: https://github.com/rajinwonderland/react-code-blocks#-demo
-import { CopyBlock as Component, irBlack } from 'react-code-blocks';
+import { useState } from 'react';
+import styled from 'styled-components';
 
 import { Token } from '../utils';
 
-export function CodeBlock({ token }: { token: Token }) {
+const CodeContainer = styled.div`
+  position: relative; /* For the copy button positioning */
+  padding: 20px;
+  background: #000;
+`;
+
+const CodeContent = styled.div`
+  color: #f8f8f8;
+  overflow-x: auto;
+  white-space: pre;
+`;
+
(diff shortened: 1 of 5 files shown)


src/components/CodeBlock.tsx
```diff
@@ -1,16 +1,82 @@
...
-      showLineNumbers={true}
-      codeBlock
-    />
+    <CodeContainer>
+      <CopyButton
+        onClick={() => {
+          handleCopy();
+        }}
+      >
+        {isCopied ? '✅' : '📋'}
```
bang9, line 77: 🤣ㅋㅋㅋ 심플하고 좋네요 이모지 짱👍👍👍

src/components/CodeBlock.tsx
```diff
@@ -1,16 +1,82 @@
...
+  height: 26px;
+  width: 26px;
+  padding: 2px;
+`;
+
+export const CodeBlock = ({ token }: { token: Token }) => {
+  const [isCopied, setIsCopied] = useState(false);
+  const handleCopy = async () => {
+    try {
+      await navigator.clipboard.writeText(code);
```
bang9, line 54: `clipboard API` ios webkit 이 살~짝 아슬해보이네용
image

bang9 (approved):
lgtm!

src/components/CodeBlock.tsx
AhyoungRyu: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard#browser_compatibility
맞아요 그부분이 조금 걸리긴 하지만.... 부가적인 기능이니 안고가려합니다 😂

PR #123, opened 2024-03-19
fix: visibility of empty elements handling

AhyoungRyu:
Addressed an issue reported in https://sendbird.slack.com/archives/C06D560RN6B/p1710827091693129

### DIFF
--- a/src/css/index.css
+++ b/src/css/index.css
@@ -1,19 +1,3 @@
-a:empty,
-ul:empty,
-dl:empty,
-div:empty,
-section:empty,
-article:empty,
-p:empty,
-h1:empty,
-h2:empty,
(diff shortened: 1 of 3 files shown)


src/css/index.css
```diff
@@ -1,19 +1,3 @@
-a:empty,
```
AhyoungRyu, line 1: In this way, there's a chance that we unintentionally override the style in user's website where the widget is being used. 

So I wrapped this config under a certain namespace; `aichatbot` which is a root component's id. https://github.com/sendbird/chat-ai-widget/blob/develop/packages/self-service/src/main.tsx#L6

bang9 (approved):

PR #126, opened 2024-03-21
[AC-1747] Enable passing ChatAiWidget props to self service widget embed code

AhyoungRyu:
There's a request from a user who's using self-service about possibility of widget customization. However, there's no way to achieve this since the embed code we provide from the dashboard is only inject appid / botid to the ChatAiWidget component and other config props are not being passed. 
So I added a connection between self-service <-> ChatAiWidget component like [this](https://github.com/sendbird/chat-ai-widget/pull/126/files#diff-8909f40095486fdc54c1782c9213f49e02cfbe73b2a7fe9295f91adef11d97c1R17).

This is a common guide we provide from Dashboard when user creates a bot.
```
<script>
      !function(w, d, s, ...args){
        var div = d.createElement('div');
        div.id = 'aichatbot';
        d.body.appendChild(div);
        w.chatbotConfig = args;
        var f = d.getElementsByTagName(s)[0],
        j = d.createElement(s);
        j.defer = true;
        j.type = 'module';
        j.src = 'https://aichatbot.sendbird.com/index.js';
        f.parentNode.insertBefore(j, f);
      }(window, document, 'script', 'AE8F7EEA-4555-4F86-AD8B-5E0BD86BFE67', 'khan-academy-bot');
</script>
```

If user wants to customize the embed code to pass the chatbot config props(to ChatAiWidget component), they can modify the above common embed code like below
```
<script>
      ... Same as above
        // This is what user needs to add; arg[0] is for appid & arg[1] is for botid 
        args[2] = {
          enableEmojiFeedback: true / false, 
          enableSourceMessage: true / false, 
          ... any other props defined in ChatAiWidget component can be listed here 
        }
      }(window, document, 'script', 'AE8F7EEA-4555-4F86-AD8B-5E0BD86BFE67', 'khan-academy-bot');
</script>
```

### DIFF
--- a/src/components/ParsedBotMessageBody.tsx
+++ b/src/components/ParsedBotMessageBody.tsx
@@ -3,10 +3,10 @@ import { ReactNode } from 'react';
 import styled from 'styled-components';
 
 import BotMessageBottom from './BotMessageBottom';
+import { CodeBlock } from './CodeBlock';
 import SourceContainer, { Source } from './SourceContainer';
 import { useConstantState } from '../context/ConstantContext';
 import { replaceWithRegex, Token, TokenType } from '../utils';
-import { CodeBlock } from './CodeBlock';
 
 const Text = styled.div`
   width: inherit;
@@ -55,7 +55,7 @@ export default function ParsedBotMessageBody(props: Props) {
   // console.log('## sources: ', sources);
   if (tokens.length > 0) {
     return (
-      <MultipleTokenTypeContainer>
+      <MultipleTokenTypeContainer className="sendbird-word">
         {tokens.map((token: Token, i) => {
           if (token.type === TokenType.string) {
             return (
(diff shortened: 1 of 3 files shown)


src/components/ParsedBotMessageBody.tsx
```diff
@@ -55,7 +55,7 @@ export default function ParsedBotMessageBody(props: Props) {
   // console.log('## sources: ', sources);
   if (tokens.length > 0) {
     return (
-      <MultipleTokenTypeContainer>
+      <MultipleTokenTypeContainer className="sendbird-word">
```
AhyoungRyu, line 58: It's a same className with UIKit. By adding this, user can customize the style of the message bubble. (It's another request from the customer 😄)

luke-cha (approved):

PR #136, opened 2024-03-27
feat: apply text & icon color variation based on theme

AhyoungRyu:
Due to the fixed text color in the error message component, the error message is not really visible in the dark theme.  So I gave different text & icon color to display a proper color on each theme(light / dark).


src/components/Chat.tsx
```diff
@@ -77,7 +77,7 @@ const SBComponent = () => {
       sdkInitParams={sdkInitParams}
       configureSession={configureSession}
       customExtensionParams={userAgentCustomParams.current}
-      breakPoint={isMobile}
```
AhyoungRyu, line 80: My bad 😅

chohongm (approved):
LGTM

PR #142, opened 2024-04-13
[CPLAT-1833] create upload_to_s3 pipeline

JooeunAhn:


.circleci/config.yml
```diff
@@ -106,9 +156,28 @@ jobs:
...
+          configure-default-region: false
+      - aws-cli/role-arn-setup:
+          profile-name: ci-web-prod
+          source-profile: default
+          role-arn: arn:aws:iam::049322569699:role/dashboard-ci-web-prod
+      - aws-s3-sync:
+          from: 'dist/'
+          to: 's3://genai-aichatbotwidget-use1-prod/'
+          arguments: '--delete'
+      - aws-create-invalidation:
```
kychool, line 174: Just in case, don't we need to run this only when there is an explicit user action, such as tagging?

kychool (approved):

.circleci/config.yml
JooeunAhn: I turned it off for pipeline testing. I turned It on

Kairo-Kim (approved):

PR #152, opened 2024-04-19
Fix incorrect feedback icon position

AhyoungRyu:
The feedback icons are next to the message sent time UI which is not a right place 

Should be positioned like below.  
Screenshot 2024-04-19 at 3 24 56 PM

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -1,4 +1,7 @@
-import { type SessionHandler, type SendbirdGroupChat } from '@sendbird/chat/lib/__definition';
+import {
+  type SessionHandler,
+  type SendbirdGroupChat,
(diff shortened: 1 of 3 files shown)


src/const.ts
```diff
@@ -1,4 +1,7 @@
-import { type SessionHandler, type SendbirdGroupChat } from '@sendbird/chat/lib/__definition';
+import {
+  type SessionHandler,
+  type SendbirdGroupChat,
+} from '@sendbird/chat/lib/__definition';
```
AhyoungRyu, line 4: Fixed by lint

chohongm (approved):
LGTM!

PR #154, opened 2024-04-19
feat: Display message data

chohongm:
https://sendbird.atlassian.net/browse/EI-919

Screenshot 2024-04-29 at 5 27 14 PM


.env
bang9, line 1: Is it okay? @AhyoungRyu 🤔

src/App.tsx
```diff
@@ -1,5 +1,6 @@
 import ChatAiWidget from './components/ChatAiWidget'; //import { ChatAiWidget } from "@sendbird/chat-ai-widget";
 import { Constant } from './const';
+import {SendbirdChatAICallbacks} from './interfaces';
```
bang9:
```suggestion
import { SendbirdChatAICallbacks } from './interfaces';
```

src/components/CustomChannelComponent.tsx
```diff
@@ -215,23 +222,26 @@ export function CustomChannelComponent() {
...
-                isLastBotMessage={isLastBotMessage}
-                messageCount={messageCount}
-              />
-              {message.messageId === lastMessage?.messageId &&
-                (dynamicReplyOptions.length > 0 ? (
-                  <DynamicRepliesPanel replyOptions={dynamicReplyOptions} />
-                ) : isStaticReplyVisible ? (
-                  <StaticRepliesPanel botUser={botUser} />
-                ) : null)}
+              <>
```
bang9: why we need fragment?

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  // console.log('## messageData: ', messageData);
+
+  const [functionCalls, setFunctionCalls] = useState<FunctionCallRenderData[]>([]);
+
+  useEffect(() => {
+    const messageDataObject: MessageDataObject = JSON.parse(messageData);
+    const functionCallsData = messageDataObject?.function_calls;
+
+    // console.log('## functionCallsData: ', functionCallsData, onViewDetailClick);
```
bang9:
```suggestion
```

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,177 @@
...
+interface FunctionCallRenderData {
+  name: string;
+  onClick: () => void;
+}
+
+export default function MessageDataContent({ messageData }: MessageDataContentProps) {
+  const { callbacks } = useConstantState();
+  const onViewDetailClick = callbacks?.onViewDetailClick;
+
+  // console.log('## messageData: ', messageData);
```
bang9:
```suggestion
```

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,177 @@
...
+              name: response.name,
+              onClick: () => functionCall(functionCallData),
+            });
+          }
+        });
+      if (newFunctionCalls.length > 0) {
+        setFunctionCalls(newFunctionCalls);
+      }
+    }
+  }, [messageData]);
```
bang9: Does `message.data` change dynamically? If not, it seems there's no reason to use useEffect and useState

bang9: `const functionCalls = getFunctionCalls(message.data);`

src/context/ConstantContext.tsx
```diff
@@ -2,12 +2,14 @@ import { LabelStringSet } from '@sendbird/uikit-react/ui/Label';
 import { createContext, useContext, useMemo } from 'react';
 
 import { type Constant, DEFAULT_CONSTANT } from '../const';
+import {SendbirdChatAICallbacks} from '../interfaces';
```
bang9:
```suggestion
import { SendbirdChatAICallbacks } from '../interfaces';
```

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,177 @@
+import styled from 'styled-components';
+
+import { useConstantState } from '../context/ConstantContext';
+import { useEffect, useState } from 'react';
+import { ReactComponent as EllipsisIcon } from '../icons/icon-ellipsis.svg';
+import { ReactComponent as MessageBubbleIcon } from '../icons/icon-message-bubble.svg';
+import { ReactComponent as ChevronRightIcon } from '../icons/icon-chevron-right.svg';
+import {ViewDetailData} from '../interfaces';
```
bang9:
```suggestion
import { ViewDetailData } from '../interfaces';
```

.env
chohongm: I assumed .env should stay local only but @AhyoungRyu  can you confirm its ok we remove it?

src/components/CustomChannelComponent.tsx
chohongm: I think I put this because Message expects single child component
```
  children?: React.ReactNode;
```
is it ok without it?

src/components/MessageDataContent.tsx
chohongm: I was thinking the same: messageData dependency is redundant.
ill remove it

src/context/ConstantContext.tsx
chohongm: let me run lint:fix

.env
AhyoungRyu: I know it's not recommended way but, the reason we're having this is we've wanted ppl to give it a try this repo without visiting the dashboard just for a trial. Let's keep this for now.

src/components/CustomChannelComponent.tsx
bang9: ReactNode type includes `Iterable<ReactNode>` ( [ReactNode, ReactNode, ReactNode, ...] )
image

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,169 @@
...
+}
+
+export default function MessageDataContent({
+  messageData,
+}: MessageDataContentProps) {
+  const { callbacks } = useConstantState();
+  const onViewDetailClick = callbacks?.onViewDetailClick;
+
+  function getFunctionCallsData() {
+    const messageDataObject: MessageDataObject = JSON.parse(messageData);
```
bang9: Could you add try/catch block?

```ts
try {
  const messagDataObject = JSON.parse(..);
} catch {
  return [];
}
```

chohongm: good point I did not consider it!
Throws:
SyntaxError – Thrown if the string to parse is not valid JSON.

src/components/CustomChannelComponent.tsx
chohongm: thanks for checking this out. i removed this!

.env
chohongm: ah gotcha. ill put it back

.gitignore
```diff
@@ -13,6 +13,7 @@ dist
 dist-ssr
 *.local
 
+.env
```
AhyoungRyu: Let's remove this one too :)

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,170 @@
...
+      const newFunctionCalls: FunctionCallRenderData[] = [];
+      if (
+        functionCallsData &&
+        Array.isArray(functionCallsData) &&
+        functionCallsData.length > 0
+      ) {
+        (functionCallsData as ViewDetailData[]).forEach(
+          (functionCallData: ViewDetailData) => {
+            const response = functionCallData.response;
+            if (response.name) {
```
AhyoungRyu: Seems it's a required field but checking if it exists. Any reason we're checking this?

src/components/CustomChannelComponent.tsx
```diff
@@ -110,8 +111,15 @@ const Root = styled.div<RootStyleProps>`
...
-    useConstantState();
+  const {
+    suggestedMessageContent,
+    botId,
+    enableEmojiFeedback,
+    customUserAgentParam,
+  } = useConstantState();
+  const isForSendbirdDashboard =
+    customUserAgentParam &&
+    customUserAgentParam['chat-ai-widget-preview'] === 'True';
```
bang9: Can you make it into a util function?

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,170 @@
...
+  const onViewDetailClick = callbacks?.onViewDetailClick;
+
+  function getFunctionCallsData() {
+    try {
+      const messageDataObject: MessageDataObject = JSON.parse(messageData);
+      const functionCallsData = messageDataObject?.function_calls;
+      const newFunctionCalls: FunctionCallRenderData[] = [];
+      if (
+        functionCallsData &&
+        Array.isArray(functionCallsData) &&
```
bang9:
```suggestion
        Array.isArray(functionCallsData) &&
```

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,170 @@
...
+        functionCallsData.length > 0
+      ) {
+        (functionCallsData as ViewDetailData[]).forEach(
+          (functionCallData: ViewDetailData) => {
```
bang9:
```suggestion
          (functionCallData) => {
```

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,170 @@
...
+            const response = functionCallData.response;
+            if (response.name) {
+              const functionCall =
+                onViewDetailClick && typeof onViewDetailClick === 'function'
+                  ? onViewDetailClick
+                  : noop;
```
bang9:
```suggestion
              const functionCall = typeof onViewDetailClick === 'function' ? onViewDetailClick : noop;
```

src/context/ConstantContext.tsx
```diff
@@ -2,12 +2,14 @@ import { LabelStringSet } from '@sendbird/uikit-react/ui/Label';
...
 
 import { type Constant, DEFAULT_CONSTANT } from '../const';
+import { SendbirdChatAICallbacks } from '../interfaces';
 
 const initialState = DEFAULT_CONSTANT;
 
 interface ConstantContextProps extends Constant {
   applicationId: string | null;
   botId: string | null;
+  callbacks?: SendbirdChatAICallbacks;
```
bang9, line 12: For DX, it would be good to align the naming conventions of **function calls** and **callbacks** interfaces to some extent.

bang9: or, please use a specific name for onViewDetailClick that is closely related to FunctionCall.

src/components/MessageDataContent.tsx
chohongm: yeah its better to validate all 
```
export interface ViewDetailResponse {
  name: string;
  response_text: string;
  status_code: number;
}
```
because it can be any data.

chohongm: ill update this

src/components/CustomChannelComponent.tsx
chohongm: like this?
```
function isForSendbirdDashboard(data: object | undefined) {
  return (
    data &&
    'chat-ai-widget-preview' in data &&
    data['chat-ai-widget-preview'] === 'True'
  );
}
```
I think redundant because used in only one place.

chohongm: updated!

src/context/ConstantContext.tsx
chohongm: I choose name `callbacks` intentionally to make it as general as possible. It should contain any callbacks for future not just for this function callback case. Currently it only has `onViewDetailClick` but it can include many more other cases in the future.

current
```
export interface SendbirdChatAICallbacks {
  onViewDetailClick?: (data: ViewDetailData) => void;
}
```
I was not sure of name `onViewDetailClick` is good or not. Do you have suggestion? @bang9

src/components/MessageDataContent.tsx
liamcho: what changed here?

src/context/ConstantContext.tsx
bang9: While handling only data related to `FunctionCall`, the naming of `ViewDetail` seems too general.
Moreover, there is no guarantee that all data, including `FunctionCall`, added in the future will be provided in the same format.
Therefore, it would be good to make the naming more specific to the current use case. (If there are common fields in the format of data added in the future, we can create a separate interface for that part.)

src/components/CustomChannelComponent.tsx
bang9: Yes, but I suggested separating and managing it separately since the key-value pairs are being managed as constants.

src/components/CustomChannelComponent.tsx
```diff
@@ -187,6 +198,8 @@ export function CustomChannelComponent() {
     return getBotWelcomeMessages(allMessages, botId);
   }, [messageCount]);
 
+  if (!channel || !botUser) return null;
```
bang9: Are there any side effects on the UI?

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,187 @@
...
+function objectOfViewDetailResponse(object: any): object is ViewDetailResponse {
+  const { name, reponse, status_code } = object;
+  return (
+    typeof name === 'string' &&
+    typeof reponse === 'string' &&
+    typeof status_code === 'number'
+  );
+}
+
+function isValidFunctionCalls(functionCallsData: object | undefined) {
```
bang9: Using the type predicate, you can eliminate the type assertion.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,187 @@
...
+    try {
+      const messageDataObject: MessageDataObject = JSON.parse(messageData);
+      const functionCallsData = messageDataObject?.function_calls;
+      const newFunctionCalls: FunctionCallRenderData[] = [];
+      if (
+        Array.isArray(functionCallsData) &&
+        functionCallsData.length > 0 &&
+        isValidFunctionCalls(functionCallsData)
+      ) {
+        (functionCallsData as ViewDetailData[]).forEach((functionCallData) => {
```
bang9: Using the type predicate, you can eliminate the type assertion.

https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/components/MessageDataContent.tsx
bang9: By using the type predicate, you can eliminate type assertions. Not only does this ensure type safety, but it also makes the code more concise and readable.

src/components/MessageDataContent.tsx
bang9: By using the type predicate, you can eliminate type assertions. Not only does this ensure type safety, but it also makes the code more concise and readable.

src/components/MessageDataContent.tsx
```diff
@@ -0,0 +1,187 @@
...
+  if (functionCalls.length === 0) return null;
+
+  return (
+    <Root>
+      <SideBar />
+      <DataContainer>
+        <DataRow>
+          <MessageBubbleIcon id="aichatbot-widget-ellipsis-icon" />
+          <Text>{}</Text>
+          <WorkFlowType>{}</WorkFlowType>
```
bang9: Do something need to be filled in here?

src/components/MessageDataContent.tsx
chohongm: nice

src/components/CustomChannelComponent.tsx
chohongm: ui will be empty. but this is safer because without it, run time error happens

src/components/MessageDataContent.tsx
chohongm: yes i committed

src/context/ConstantContext.tsx
chohongm: Can you provide an example for the naming?

chohongm: callbacks has onXxxClick convention so i think name looks ok but ViewDetailData -> FunctionCallData looks better

src/components/CustomChannelComponent.tsx
AhyoungRyu: Isn't it okay without this in develop branch? Please check this one before you merge 👍

package-lock.json
AhyoungRyu, line 1: It seems the lock file is changed unintentionally. Please check this one too before you merge.

AhyoungRyu (approved):

src/components/CustomChannelComponent.tsx
chohongm: I added for safety but not tested. ill put back

src/context/ConstantContext.tsx
chohongm: updated
```
export interface SendbirdChatAICallbacks {
  onViewDetailClick?: (data: FunctionCallData) => void;
}
```

src/hooks/useSendMessage.ts
```diff
@@ -9,6 +9,7 @@ export function useSendMessage(): (message: string) => Promise<void> {
       if (currentChannel == null) {
         throw new Error('currentChannel is not defined');
       } else {
+        console.log('## here');
```
bang9:
```suggestion
```

bang9 (approved):

PR #158, opened 2024-04-22
fix: feedback diglog position

AhyoungRyu:
**AS-IS**
Screenshot 2024-04-22 at 2 30 07 PM

**TO-BE**
The dialog should be attached to the bottom of widget window element not the entire DOM. 
Screenshot 2024-04-22 at 2 19 10 PM

### DIFF
--- a/src/hooks/useDynamicAttachModal.tsx
+++ b/src/hooks/useDynamicAttachModal.tsx
@@ -1,0 +1,37 @@
+import { useEffect } from 'react';
+
+import { elementIds } from '../const';
+
+const parentId = elementIds.widgetWindow;
+const childId = elementIds.uikitModal;
+
+function useDynamicAttachModal() {
+  useEffect(() => {
+    // Set up a MutationObserver to monitor changes in the parent element
+    const observer = new MutationObserver((mutations) => {
+      mutations.forEach((mutation) => {
+        if (mutation.addedNodes.length) {
+          // Check if the added nodes include the one with parentId
+          Array.from(mutation.addedNodes).forEach((node) => {
+            if (node?.id === parentId) {
+              const parent = document.getElementById(parentId);
+              const child = document.getElementById(childId);
+              if (parent && child) {
+                parent.appendChild(child);
+              }
+            }
+          });
+        }
+      });
+    });
+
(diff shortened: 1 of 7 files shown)


src/hooks/useDynamicAttachModal.tsx
```diff
@@ -0,0 +1,37 @@
...
+                parent.appendChild(child);
+              }
+            }
+          });
+        }
+      });
+    });
+
+    // Set the observer to monitor the entire document
+    observer.observe(document.body, { childList: true, subtree: true });
```
bang9, line 31: If attached to the body, it seems there may be overhead detecting changes to the entire DOM unrelated to the widget. How about testing it by attaching it to 'sendbird-modal-root'?

bang9: Oh, so it's not about attaching individual modals added but moving the entire modal root under the widget window.

AhyoungRyu: Yeah that's correct.

bang9: After moving the modal root, if you disconnect it, will the modal root return to its original position during the next rendering?

AhyoungRyu: Hmm it'd be better if there's a way to get a root element of this widget module than using `document.body`. 

>After moving the modal root, if you disconnect it, will the modal root return to its original position during the next rendering?

No it won't. It'll be under `id: aichatbot-widget-window` element once it's moved. Do you think it should be put back to its original position? @bang9

bang9 (approved):

src/hooks/useDynamicAttachModal.tsx
AhyoungRyu: @bang9 As we discussed in Slack, I changed it to disconnect right after it moves the modal root element to the widget window element. bd19db95ba94bd705f01d2d9b2832cc4cf99e93f
Also, it's working as expected in both mobile & desktop view. 

- Desktop
Screenshot 2024-04-22 at 5 46 52 PM

- Mobile
Screenshot 2024-04-22 at 5 47 07 PM

PR #169, opened 2024-04-24
feat: Display file message from bot

chohongm:
[수천만원에 엑싯하는 사이드 프로젝트가 한둘이 아니다 (feat. 비즈니스 마켓플레이스) (2).webm](https://github.com/sendbird/chat-ai-widget/assets/16806397/7a6a5ee7-35d9-4266-8783-8c701dda17fa)

Note: empty message above is expected (empty user message was sent before file message)

File viewer will be added in the future (currently no design so I left it as comment instead)
[수천만원에 엑싯하는 사이드 프로젝트가 한둘이 아니다 (feat. 비즈니스 마켓플레이스).webm](https://github.com/sendbird/chat-ai-widget/assets/16806397/52221ad2-ec5c-445b-aebe-77560e822a2f)


src/components/FileMessage.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const {
+    isScrollBottomReached,
+    scrollToBottom,
+  } = useGroupChannelContext();
+
+  // const root = document.getElementById('aichatbot-widget-window');
+
+  return (
+    <Root>
+      {/*{root &&*/}
```
AhyoungRyu: Do we want to keep this?

src/utils/index.ts
```diff
@@ -255,3 +254,14 @@ export const localStorageHelper = () => {
...
+
+export function downloadFileWithUrl(url?: string | null) {
+  let safeURL = url;
+  if (safeURL) {
+    if (!safeURL.startsWith('http://') && !safeURL.startsWith('https://')) {
+      safeURL = 'https://' + safeURL;
+    }
+    window.open(safeURL, '_blank', 'noopener,noreferrer');
+  }
+  return Promise.resolve();
```
AhyoungRyu: Any reason we return this?

chohongm: its just dummy code because the prop expects return type to be Promise<void>

src/components/CustomMessage.tsx
```diff
@@ -58,12 +58,12 @@ export default function CustomMessage(props: Props) {
...
 
   // admin message
-  if (message.messageType === 'admin') {
+  if (message.isAdminMessage()) {
     return <div>{<AdminMessage message={message} />}</div>;
   }
 
   if (isFormMessage(message)) {
-    const forms = message.extendedMessagePayload.forms;
+    const forms: Form[] = message.extendedMessagePayload!.forms as Form[];
```
AhyoungRyu, line 66: Any reason we force type casting here in this way?

src/components/FileMessage.tsx
chohongm: yeah I prefer to keep this because I highly think we will need to add Fileviewer in the future. It will be easier to use this then

src/components/localhost:5173/chat-ai-widget/src/components/CustomMessage.tsx
AhyoungRyu, line 1: I guess this file has been added accidentally :)

src/components/FileMessage.tsx
AhyoungRyu: Then let's add a brief comment about it

src/components/CustomMessage.tsx
chohongm: because forms is unknown type so `forms[0]` complains

bang9: ~~Is it different? `message.forms` vs `message.extendedMessagePayload.forms`~~

src/components/CustomMessage.tsx
```diff
@@ -1,14 +1,14 @@
 import { User } from '@sendbird/chat';
-import { UserMessage } from '@sendbird/chat/message';
+import { BaseMessage, UserMessage } from '@sendbird/chat/message';
 // eslint-disable-next-line import/no-unresolved
-import { EveryMessage } from 'SendbirdUIKitGlobal';
```
bang9, line 4: 👍 could you remove `// eslint-disable-next-line import/no-unresolved` too?

src/components/FileMessage.tsx
bang9: tip. you can comment out jsx like this

```
{/*<LongPropsComponents
    ...
    ...
    ...
    ...
    ...
    ...
/>*/}
```

src/utils/index.ts
bang9: tip.

```ts
function downloadFileWithUrl() {
  // ...
  return Promise.resolve();
}

// ⬇️⬇️⬇️

async function downloadFileWithUrl() {
  // ...
}
```

src/utils/messages.ts
```diff
@@ -82,7 +82,7 @@ export function getBotWelcomeMessages(
 }
 
 export function isFormMessage(message: EveryMessage) {
```
bang9: tip. by using a type predicate, you don't need to perform type assertion `as Form[]`.

```suggestion
export function isFormMessage(message: EveryMessage): message is EveryMessage & { extendedMessagePayload: { forms: Form[] }} {
```

src/components/FileMessage.tsx
```diff
@@ -0,0 +1,59 @@
...
+// import {createPortal} from 'react-dom';
+
+const Root = styled.div`
+  width: 100%;
+`;
+
+const Image = styled.img`
+  border-radius: 16px;
+  width: 100%;
+`;
```
bang9: It seems like these simple components don't need to use styled components.
Since styled components are executed at runtime, they need to be executed, and template literals not to be minified, making the code larger.

How about just using inline styles on regular tags?

src/utils/index.ts
chohongm: @bang9 ah that is better thanks ill appy this

src/components/FileMessage.tsx
chohongm: > Since styled components are executed at runtime, they need to be executed, and template literals not to be minified, making the code larger.

Thats a good point. Additionally,I learned that using styled component has no css class identifier which has two crucial cons:
- Hard to locate source code from developer tools.
- Customers cannot override their own style.

I think we should remove styled component later. For the same reason, I want to avoid using inline styling. Also inline styling has high precedence so customer may complain later. How about using scss files just like UIKit?

bang9: Agreed. However, the highest priority in the current widget is to improve performance and reduce bundle size.

bang9: Later on, it would be good to transition to zero-runtime tools like vanilla extract.

- https://linaria.dev/
- https://stylexjs.com/
- https://vanilla-extract.style/

bang9: https://youtu.be/9JZHodNR184?si=tw6eOb_pxqsHyEt9&t=220

src/components/CustomMessage.tsx
bang9: maybe now, we can remove assertion (`as Form[]`)

bang9 (approved):

AhyoungRyu:
From the video you attached, I just saw this is being rendering in the image viewer message. Please make sure this is not being rendered before you merge :) 

Screenshot 2024-04-26 at 6 10 30 PM


>Note: empty message above is expected (empty user message was sent before file message)

Oh probably this is what you mentioned.

PR #175, opened 2024-04-25
Revert "Put datadog-rum back with silm version 

AhyoungRyu:
We've decided not to have Datadog related logic in Widget. So removing all related code.


bang9 (approved):

bang9:
byebye 🐶

PR #180, opened 2024-04-29
[AC-1971]: Use git submodule for uikit-react to reduce the bundle size

AhyoungRyu:
### Key Changes:

- Removed the `@sendbird/uikit-react` dependency from `package.json`.
- Linked the necessary code from [sendbird/uikit-react](https://github.com/sendbird/sendbird-uikit-react) directly into the `packages/uikit/` dir through a Git submodule. <- See `.gitmodules` file
- Updated our build process(Github workflow / Circleci config) to initialize and update the Git submodule, ensuring that the latest version of the code is always used.

So how much of the bundle size would be reduced? 
**AS-IS**
Screenshot 2024-04-29 at 5 55 49 PM

**TO-BE**
Screenshot 2024-04-29 at 5 50 40 PM


package.json
```diff
@@ -14,6 +15,7 @@
     "url": "https://github.com/sendbird/chat-ai-widget.git"
   },
   "scripts": {
+    "install:deps": "npm install --force && git submodule update --init",
```
AhyoungRyu: Why we're using `--force` param here? There's a peer dep setting in `@mojojoejo/vite-plugin-purgecss` like below, so we can't use the latest vite version. So I put `--force` option to ignore the error.
Screenshot 2024-04-29 at 5 56 51 PM

src/components/BotProfileImage.tsx
```diff
@@ -1,7 +1,7 @@
 import styled from 'styled-components';
 
 import { getColorBasedOnSaturation } from '../colors';
-import { ReactComponent as Icon } from '../icons/bot-profile-image-small.svg';
```
AhyoungRyu, line 4: Why this format have to be changed? 
- Verbose to use
- It causes a conflict with UIKit svg import type and throwing an error 🥲

package.json
bang9: Could you test if it also works using the Rollup plugin instead of the Vite plugin?
- https://v1.purgecss.com/with-rollup

AhyoungRyu: @bang9  Hmm it seems it also conflicts with other packages and doesn't give a happy result. Let me split this purgecss task and do it separately :) 

I'll put the purgecss related stuff.

bang9: 😭

bang9: If the reduction in bundle size for each Vite version is not significantly noticeable, it might be better to maintaining version compatibility

AhyoungRyu: Yeah it seems there's no diff between 5 and 6 in terms of the bundle size. So I just used the latest version of Vite which is 5.2.10 and updated other plugins' version accordingly. https://github.com/sendbird/chat-ai-widget/pull/180/commits/124f51ec746820db2747e3f4bb9e7858bd1e7dbd

.eslintrc.cjs
```diff
@@ -2,38 +2,40 @@ module.exports = {
   root: true, // This fixes issue when running lint fix command: https://stackoverflow.com/questions/55060228/eslint-couldnt-find-the-plugin-eslint-plugin-typescript-eslint
   env: { browser: true, es2020: true },
   parser: '@typescript-eslint/parser',
-  parserOptions: { ecmaVersion: 'latest', sourceType: 'module' },
-  plugins: ['import', 'styled-components-a11y'],
+  parserOptions: {
+    ecmaVersion: 'latest',
+    sourceType: 'module',
+    ecmaFeatures: { jsx: true },
```
chohongm, line 8: jsx: ECMScript 규격의 [JSX](https://facebook.github.io/jsx/) 사용 여부
^ is this relevant to reducing bundle size?

.eslintrc.cjs
```diff
@@ -2,38 +2,40 @@ module.exports = {
...
+  },
+  plugins: ['import', '@typescript-eslint', 'styled-components-a11y'],
```
chohongm: why we need @typescript-eslint?

src/components/MessageDataContent.tsx
```diff
@@ -1,9 +1,9 @@
 import styled from 'styled-components';
 
 import { useConstantState } from '../context/ConstantContext';
-import { ReactComponent as ChevronRightIcon } from '../icons/icon-chevron-right.svg';
-import { ReactComponent as EllipsisIcon } from '../icons/icon-ellipsis.svg';
-import { ReactComponent as MessageBubbleIcon } from '../icons/icon-message-bubble.svg';
```
chohongm, line 6: im not sure why we had to use
{ ReactComponent as xxx } format in the first place

tsconfig.json
```diff
@@ -20,7 +20,10 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "noFallthroughCasesInSwitch": true,
-    "allowSyntheticDefaultImports": true
+    "allowSyntheticDefaultImports": true,
+    "paths": {
+      "@uikit/*": ["./packages/uikit/src/*"],
+    }
```
chohongm, line 26: what is path for?

vite.config.ts
```diff
@@ -12,12 +18,29 @@ export default defineConfig({
...
+      },
+    }),
     dts(),
+    visualizer({
+      filename: './dist/report.html',
+      brotliSize: true,
+    }),
+    // pluginPurgeCss({
+    //   variables: true,
+    // }),
```
chohongm: lets remove unused

vite.config.ts
```diff
@@ -12,12 +18,29 @@ export default defineConfig({
...
   ],
-  define: {
-    APP_VERSION: JSON.stringify(process.env.npm_package_version),
+  resolve: {
+    alias: [
+      {
+        find: '@uikit',
+        replacement: path.resolve(__dirname, 'packages/uikit/src'),
+      },
+    ],
```
chohongm, line 39: what is resolve for?

vite.config.ts
```diff
@@ -26,19 +49,17 @@ export default defineConfig({
...
       formats: ['es', 'umd'],
-      fileName: (format) => `index.${format}.js`
+      fileName: (format) => `index.${format}.js`,
     },
     rollupOptions: {
-      external: [
-        'react',
-        'react-dom',
-      ],
+      plugins: [terser()],
```
chohongm, line 51: great idea! for curiousity, how much in % is size reduced by terser?

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "packages/uikit"]
+	path = packages/uikit
+	url = git@github.com:sendbird/sendbird-uikit-react.git
+	branch = ai-widget/experimental
```
chohongm, line 4: what is this file for?

chohongm (commented):
Is terser the only change that reduced bundle size?

tsconfig.json
AhyoungRyu: Related to the thing I mentioned in the key change. 

>what is path for?
 
To enable this `@uikit/***`.

vite.config.ts
AhyoungRyu: Also related to https://github.com/sendbird/chat-ai-widget/pull/180#discussion_r1584204401

.gitmodules
AhyoungRyu: Related to the things I mentioned in the PR description. 
Screenshot 2024-04-30 at 3 39 04 PM

Let me know if you need more detailed explanation :)

tsconfig.json
chohongm: so this is necessary when using submodule?

.eslintrc.cjs
AhyoungRyu: Size reduce related? Not really but added in terms of better practice? \cc @bang9

vite.config.ts
AhyoungRyu: - With
Screenshot 2024-04-30 at 3 44 31 PM

- Without
Screenshot 2024-04-30 at 3 44 05 PM

AhyoungRyu:
>Is terser the only change that reduced bundle size?

Yes :)

vite.config.ts
AhyoungRyu: I commented this out since we're talking about the purgecss in here https://sendbird.slack.com/archives/C05QD5M7XDH/p1714457280742949?thread_ts=1714383967.760279&cid=C05QD5M7XDH. TBD yet 🤔

.eslintrc.cjs
chohongm: before adding this was there an issue?

vite.config.ts
chohongm: oh it is less than what I expected (>= 20%)

chohongm (approved):
LGTM! Great job~

.eslintrc.cjs
AhyoungRyu: It seems we don't need to specify this in both `plugins` and `extends`. So I removed it from plugins https://github.com/sendbird/chat-ai-widget/pull/180/commits/dccade2231ce2c82258e93fbf8fbdaae43e576ae

.eslintrc.cjs
bang9: I directly linked UIKit and set up path aliases, ensuring that files are processed correctly in ESLint as well.
But I can't remember why `ecmaFeatures: { jsx: true }` was enabled. 🙄

package.json
```diff
@@ -29,37 +31,45 @@
...
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-styled-components-a11y": "^2.1.32",
     "prettier": "2.8.8",
+    "prettier-eslint": "^15.0.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "rollup-plugin-terser": "^7.0.2",
+    "rollup-plugin-visualizer": "^5.12.0",
     "tsc-silent": "^1.2.2",
+    "tslint-config-prettier": "^1.18.0",
```
bang9: We are using eslint-typescript instead of the deprecated tslint. It seems like we can remove it.

bang9 (approved):

package.json
AhyoungRyu: Yeah removed it too https://github.com/sendbird/chat-ai-widget/pull/180/commits/2d02b80d533f7e74659d7736f05f9321f7ae5877

vite.config.ts
AhyoungRyu: Removed it https://github.com/sendbird/chat-ai-widget/pull/180/commits/26d1e7283ae06399b12a079cd4ce2e8d2bb02f13 :)

PR #185, opened 2024-04-30
[AC-2203] feat: support dynamic service name

bang9:
- Added serviceName to chatbot configs
  - [service name list and policy](https://docs.google.com/spreadsheets/d/187QwJpyBzAQGFIb-_slunuixu8yI6BbOnr1dwTpEsRo/edit#gid=0)

The service names are as follows. When injecting into plugins that want to be tracked separately, you should inject serviceName.

- If using npm, `genai-chatbot-widget` is used.
- If used as an embedding in a general web environment, `genai-self-service` is used.
- If used as an embedding in a WordPress environment, `genai-wordpress-self-service` is used.
- If used as a plugin in a WordPress environment, `genai-wordpress-plugin` is used.
- If used as an embedding in a Shopify environment, `genai-shopify-self-service` is used.
- If used as a plugin in a Shopify environment, `genai-shopify-plugin` is used.

### DIFF
--- a/packages/self-service/src/App.tsx
+++ b/packages/self-service/src/App.tsx
@@ -5,32 +5,28 @@ import {
   ChatAiWidgetConfigs,
 } from '@sendbird/chat-ai-widget';
 
-type ChatbotConfig = Window &
-  typeof globalThis & {
-    chatbotConfig: string[];
-  };
-
-const APP_ID = (window as ChatbotConfig).chatbotConfig?.[0];
-const BOT_ID = (window as ChatbotConfig).chatbotConfig?.[1];
-const chatbotConfigs =
+type AppId = string;
+type BotId = string;
+type ChatbotWindow = typeof window & {
   // Available configs are defined in the ChatAiWidget component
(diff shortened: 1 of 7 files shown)


packages/self-service/src/App.tsx
```diff
@@ -5,32 +5,28 @@ import {
...
+type AppId = string;
+type BotId = string;
+type ChatbotWindow = typeof window & {
   // Available configs are defined in the ChatAiWidget component
-  ((window as ChatbotConfig)
-    .chatbotConfig?.[2] as unknown as ChatAiWidgetConfigs) ?? {};
+  chatbotConfig?: [AppId, BotId, ChatAiWidgetConfigs];
+};
+
+const [appId, botId, configs] = (window as ChatbotWindow).chatbotConfig ?? [];
```
AhyoungRyu, line 16: 👍

AhyoungRyu (approved):

AhyoungRyu (approved):

PR #190, opened 2024-04-30
fix: fixed useWidgetButtonActivityTimeout

bang9:
- Fixed a disconnect ws issue

ticket: [AC-2225]

[AC-2225]: https://sendbird.atlassian.net/browse/AC-2225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/hooks/useChannelStyle.ts
+++ b/src/hooks/useChannelStyle.ts
@@ -101,6 +101,7 @@ export const useChannelStyle = () => {
         console.error(error);
         return {
           ...DEFAULT_CHANNEL_STYLE,
+          autoOpen: false,
           isError: true,
           errorMessage: (error as Error)?.message,
         };
@@ -109,7 +110,7 @@ export const useChannelStyle = () => {
   });
   const fetching = isPending || isLoading || isFetching;
   return useMemo(() => {
-    if (data == null) return { ...DEFAULT_CHANNEL_STYLE, isFetching: fetching };
+    if (data == null) return { ...DEFAULT_CHANNEL_STYLE, autoOpen: false, isFetching: fetching };
     return { ...data, isFetching: fetching };
   }, [data != null, fetching]);
 };
--- a/src/hooks/useChannelStyle.ts
+++ b/src/hooks/useChannelStyle.ts
@@ -9,10 +9,22 @@ import { useConstantState } from '../context/ConstantContext';
 import { isPastTime, resolvePath } from '../utils';
 
 const DEFAULT_CHANNEL_STYLE = {
-  theme: 'dark',
+  theme: 'light' as 'light' | 'dark',
   accentColor: '#742DDD',
   botMessageBGColor: '#EEEEEE',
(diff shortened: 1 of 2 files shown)


src/hooks/useWidgetButtonActivityTimeout.ts
```diff
@@ -13,49 +13,45 @@ const WS_IDLE_TIMEOUT = 6000 * 3;
...
 
     return () => {
       button.removeEventListener('click', handleClick);
-      if (timerRef.current) {
-        clearTimeout(timerRef.current);
-      }
+      clearTimer();
     };
-  }, [sdk?.reconnect]);
+  }, [sdk, initialized, channelStyle?.autoOpen]);
```
AhyoungRyu, line 54: Putting sdk(an object) will trigger re-rendering whenever its reference changes. Probably we'd better pick a value of it like `sdk?.currentUser?.userId` . What do you think?

src/hooks/useChannelStyle.ts
```diff
@@ -109,7 +110,7 @@ export const useChannelStyle = () => {
   });
   const fetching = isPending || isLoading || isFetching;
   return useMemo(() => {
-    if (data == null) return { ...DEFAULT_CHANNEL_STYLE, isFetching: fetching };
+    if (data == null) return { ...DEFAULT_CHANNEL_STYLE, autoOpen: false, isFetching: fetching };
```
AhyoungRyu: Shall we replace DEFAULT_CHANNEL_STYLE like below so `autoOpen: false` can be included? 
```
const DEFAULT_CHANNEL_SETTING = {
    theme: 'light', // not dark which is my bad 😂 
    accentColor: '#742DDD',
    botMessageBGColor: '#EEEEEE',
    autoOpen: false,
}

src/hooks/useWidgetButtonActivityTimeout.ts
bang9: The dependency array should contain references that we want to trigger side effects.
For the logic where we want to trigger side effects, I believe just the `sdk` is sufficient, as we can only execute it if the `sdk` instance exists and is `initialized`. (Since `sdk` is a class instance managed by state, its reference doesn't change unless we actually create a new `sdk` instance.)

src/hooks/useChannelStyle.ts
bang9: Good~ addressed in https://github.com/sendbird/chat-ai-widget/pull/190/commits/af6972020f331ac360fc5c39bb262b53d24fd8ae

chohongm (approved):
Nice catch LGTM!

bang9:
Actual issue is here: https://github.com/sendbird/chat-ai-widget/pull/192/commits/98c23ee2a284e3064e06ede23bb3f642b2789c44

PR #193, opened 2024-05-03
[AC-2230] fix: update types

bang9:
[AC-2230]

[AC-2230]: https://sendbird.atlassian.net/browse/AC-2230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/CustomMessage.tsx
```diff
@@ -75,19 +80,25 @@ export default function CustomMessage(props: Props) {
...
         )}
       </div>
     );
   }
 
   // Sent by other users
-  if (message.isUserMessage() && message.sender?.userId !== botUser.userId) {
+  if (
+    message.isUserMessage() &&
+    getSenderUserIdFromMessage(message) !== (botUser?.userId ?? '##bot_id##')
```
bang9: Since both the left and right values ​​can be undefined, I have inserted a magic constant for the botUser id.

chohongm: Unlikely but there can be other member with userId "##bot_id##". When can `getSenderUserIdFromMessage(message)` be undefined? I think all messages have sender info unless admin message but admin message is returned before so its ok?

src/const.ts
```diff
@@ -53,12 +54,13 @@ export const DEFAULT_CONSTANT: Constant = {
     width: '16px',
     height: '16px',
     onClick: noop,
+    style: {},
```
chohongm, line 57: where is this from?

src/custom.d.ts
```diff
@@ -1,13 +1,5 @@
...
-  const src: string;
-  export default src;
-}
-
-declare module '@sendbird/chat-ai-widget' {
-  import ChatAiWidget = require('@sendbird/chat-ai-widget');
-  const ChatAiWidget: ChatAiWidget.ChatAiWidget;
-  const Chat: ChatAiWidget.Chat;
-  export { ChatAiWidget, Chat };
+  export default ReactComponent;
```
chohongm, line 4: Could you explain this file change? I do not understand the removed nor added

chohongm (approved):
LGTM

src/custom.d.ts
bang9: This changes from https://github.com/sendbird/chat-ai-widget/pull/180#discussion_r1582787203
It just type definition for svg import

```ts
// Before
import xxx from 'xxx.svg'; // string type

// After
import xxx from 'xxx.svg'; // ReactComponent type
```

src/const.ts
bang9: It just default value

src/components/CustomMessage.tsx
bang9: make sense! let me remove `##bot_id##`

src/const.ts
chohongm: i meant there is no property called style in `Constant`

src/custom.d.ts
chohongm: we dont need 
```
declare module '@sendbird/chat-ai-widget' {	
  import ChatAiWidget = require('@sendbird/chat-ai-widget');	
  const ChatAiWidget: ChatAiWidget.ChatAiWidget;	
  const Chat: ChatAiWidget.Chat;	
  export { ChatAiWidget, Chat };
```
?

src/const.ts
bang9: It's a `Constant.customRefreshComponent.style`

src/custom.d.ts
bang9: Yes, it seems to have been declared for use under `packages/*`, but it should not be used this way (package should be install to node_modules properly.)

src/const.ts
chohongm: oh my bad!!

chohongm (commented):
how about adding 
```
export default function ChatAiWidget(props: ChatAiWidgetProps) {
  const { applicationId, botId} = props;
  if (!applicationId || !botId) {
    console.log('applicationId and botId must be non empty string values: ', applicationId, botId);
  }
  return (
    <ProviderContainer {...props}>
      {isMobile ? <MobileComponent /> : <DesktopComponent />}
    </ProviderContainer>
  );
}
```
?

PR #195, opened 2024-05-03
bug-fix: Fix scroll bottom not working with last message with suggested replies

chohongm:
### Changelogs
- Fixed a bug where initial scroll position is not at the bottom when last message has suggested replies
- Fixed a bug where margin between neighboring messages is not always `16px`

### DIFF
--- a/src/components/CustomMessage.tsx
+++ b/src/components/CustomMessage.tsx
@@ -31,7 +31,6 @@ type Props = {
   message: CoreMessageType;
   activeSpinnerId: number;
   botUser?: User;
-  lastMessageRef: React.RefObject<HTMLDivElement>;
   isBotWelcomeMessage: boolean;
   isLastBotMessage: boolean;
   messageCount: number;
@@ -44,7 +43,6 @@ export default function CustomMessage(props: Props) {
     message,
     activeSpinnerId,
     botUser,
-    lastMessageRef,
     chainTop,
     chainBottom,
     isBotWelcomeMessage,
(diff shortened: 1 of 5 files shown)


src/components/CustomMessage.tsx
```diff
@@ -100,7 +98,7 @@ export default function CustomMessage(props: Props) {
     getSenderUserIdFromMessage(message) !== botUser?.userId
   ) {
     return (
-      <div ref={lastMessageRef}>
+      <div>
```
AhyoungRyu: Since we added a new div in the parent, I think we can remove this one.

src/components/CustomMessage.tsx
```diff
@@ -154,7 +152,7 @@ export default function CustomMessage(props: Props) {
     });
 
     return (
-      <div ref={lastMessageRef}>
+      <div>
```
AhyoungRyu: This one too.

src/components/ProviderContainer.tsx
```diff
@@ -105,6 +112,11 @@ const SBComponent = ({ children }: { children: React.ReactElement }) => {
                 },
                 enableVoiceMessage: false,
                 enableFeedback: enableEmojiFeedback,
+                /**
+                 * Please keep below comments for easy local testing.
+                 */
+                // enableSuggestedReplies: true,
+                // showSuggestedRepliesFor: 'all_messages',
```
AhyoungRyu: We're using a custom suggested reply component in Widget at the moment, so this won't work 😅

AhyoungRyu (approved):

src/components/ProviderContainer.tsx
chohongm: removed it!

PR #199, opened 2024-05-07
[AC-2255] Fix incorrect feedback modal position

AhyoungRyu:
Fixes https://sendbird.atlassian.net/browse/AC-2255


src/css/index.css
```diff
@@ -91,3 +91,8 @@ input:focus {
   display: flex;
   gap: 4px;
 }
+
+.sendbird-modal-root {
+  position: absolute;
+  z-index: 2147483647;
```
chohongm, line 97: can we replace with MAX_Z_INDEX?

chohongm (approved):
Nice job :) LGTM

src/css/index.css
AhyoungRyu: Oh we can't import the module(js file; const.ts) in .css file 🥲

PR #200, opened 2024-05-07
chore: remove packages/url-webdemo

AhyoungRyu:
Since packages/url-webdemo has been moved to https://github.com/sendbird/ai-bot-url-webdemo, removing the all unused code.


.github/workflows/gh-deploy.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and Deploy
```
AhyoungRyu, line 1: Modified the confusing workflow name too.

bang9 (approved):

PR #203, opened 2024-05-08
[AC-2233] chore: add message filter for business requirements

bang9:
- When connected to the Salesforce connector, there are cases where unintended admin messages appear due to system messages. We are filtering these out.

---
## Before
image

- 불필요한 어드민 메세지가 표기됨 (어드민 메세지로 인해서 suggested replies 가 표시되지 않음)

## After
image

- 불필요한 어드민 메세지 제거 & suggested replies 표시

ticket: [AC-2233]

[AC-2233]: https://sendbird.atlassian.net/browse/AC-2233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/messages.ts
+++ b/src/utils/messages.ts
@@ -187,7 +187,7 @@ const messageFilter = {
     return message.message === "The channel's custom_type was updated.";
   },
 };
-export function shouldFilterMessage(message: BaseMessage) {
+export function shouldFilterOutMessage(message: BaseMessage) {
   if (message.isAdminMessage()) {
     return messageFilter.isSystemMessageFromSalesforce(message);
   }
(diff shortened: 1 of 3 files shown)


src/utils/messages.ts
```diff
@@ -181,3 +181,17 @@ export function getSenderUserIdFromMessage(
   // @ts-ignore
   return message?.sender?.userId ?? undefined;
 }
+
+const msgFilter = {
+  sys: {
```
AhyoungRyu, line 186: Q. Do we have reason for having `sys` here?

bang9: I wanted to express that I'm filtering messages sent as system messages.
Using "systemMessage" as the key increases the module size, so I wrote it as "sys" instead ㅜ0ㅜ

AhyoungRyu: 🤦‍♀️

AhyoungRyu (approved):

PR #222, opened 2024-05-17
[AC-2366] fix: move target if styled-components is affected by external manipulation

bang9:
When directly manipulating the head tag as shown below, there is an issue where styled-components lose their context. In the event that initialized styled-components are re-added, they are considered as directly manipulated into the head tag, prompting the target node to be moved to the body.

```ts
document.head.innerHTML += '';
```

ticket: https://sendbird.atlassian.net/browse/AC-2366
history: https://sendbird.slack.com/archives/C06TL3TJGF8/p1715934439470889


chohongm (approved):
LGTM!

PR #226, opened 2024-05-20
feat: use yarn 4

bang9:
- drop npm and use yarn@4


chohongm (approved):
LGTM. just one quesut:

after yarn install i get
Screenshot 2024-05-20 at 3 15 50 PM
can these be ignored?

the diff is saying binary files are dffierent

diff --git a/.yarn/install-state.gz b/.yarn/install-state.gz
index 2e212b47..fa70c0d1 100644
Binary files a/.yarn/install-state.gz and b/.yarn/install-state.gz differ

[it seems we should not commit this](https://yarnpkg.com/getting-started/qa). Lets add it to gitignore

- .yarn/install-state.gz
- .pnp.*

AhyoungRyu (approved):
👍

PR #228, opened 2024-05-20
ci: enable corepack

bang9:
- enable corepack


chohongm (approved):
Nice! so we need this for using yarn in image

PR #231, opened 2024-05-21
feat: Added onWidgetExpandStateChange

chohongm:
Requested by dashoard to add `onWidgetExpandStateChange`

### Why they need?
- “expand / collapse 를 그대로 면 대시보드 상에서 디자인이 깨지는데 이 때 상태 맞춰서 css 오버라이드를 해줘야 하는 경우도 있어서입니다“
Their style breaks when expanded state changes. So they need callback so they can change their css accordingly.

### Changelog
- Added `onWidgetExpandStateChange` to `Constant`

### DIFF
--- a/src/components/WidgetWindow.tsx
+++ b/src/components/WidgetWindow.tsx
@@ -100,13 +100,13 @@ const StyledCloseButton = styled.button`
 const WidgetWindow = ({ children }: { children: React.ReactNode }) => {
   const { isOpen, setIsOpen } = useWidgetOpen();
   const [isExpanded, setIsExpanded] = useState(false);
-  const { customUserAgentParam, onWidgetExpandStateChange } =
+  const { customUserAgentParam, callbacks } =
     useConstantState();
 
   const onExpandButtonToggle = () => {
     setIsExpanded((prev) => {
       const newIsExpanded = !prev;
-      onWidgetExpandStateChange?.(newIsExpanded);
+      callbacks?.onWidgetExpandStateChange?.(newIsExpanded);
       return newIsExpanded;
     });
   };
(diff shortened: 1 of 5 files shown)


src/components/WidgetWindow.tsx
```diff
@@ -100,7 +100,13 @@ const StyledCloseButton = styled.button`
...
   const { isOpen, setIsOpen } = useWidgetOpen();
   const [isExpanded, setIsExpanded] = useState(false);
-  const { customUserAgentParam } = useConstantState();
+  const { customUserAgentParam, onWidgetExpandStateChange } =
+    useConstantState();
+
+  const onExpandButtonToggle = () => {
+    onWidgetExpandStateChange?.(!isExpanded);
+    setIsExpanded((prev) => !prev);
+  };
```
chohongm, line 112: Note that below solution redundantly calls twice
https://sendbird.slack.com/archives/C024Q3BGCRL/p1716261091908299

chohongm: turned out react strictmode calls hook twice.

chohongm: Since strictmode is used in dev only, at production there wont be redundancy issue with above solution. Changed to above solution.

src/const.ts
```diff
@@ -155,6 +155,7 @@ export interface Constant extends Required<ConstantProps> {
   botStudioEditProps?: BotStudioEditProps;
   widgetOpenState?: boolean;
   onWidgetOpenStateChange?: (params: OnWidgetOpenStateChangeParams) => void;
+  onWidgetExpandStateChange?: (isExpanded: boolean) => void;
```
AhyoungRyu: Could you rebase from the develop branch and mark this one like others?
```
  /**
   * @private Callback to be called when the widget expand state changes.
   */
  onWidgetExpandStateChange?: (isExpanded: boolean) => void;
```

chohongm: nice!

chohongm: done!

bang9 (approved):

AhyoungRyu (approved):

PR #240, opened 2024-05-23
refactor: configureSession as params

bang9:
In chat SDK, because of the instance check in SessionHandler,
customer cannot use SessionHandler when using self-service or umd builds.

Therefore, we are refactoring it to also handle it as a general object.

example
```ts
const chatbotConfigs = {};

chatbotConfigs.userId = 'USER_ID';
chatbotConfigs.configureSession = () => ({
    onSessionTokenRequired: (resolve, reject) => {
      issueSessionToken('USER_ID')
        .then(token => resolve(token));
        .catch(err => reject(err));
    },
    onSessionRefreshed: () => {
      // Handle session refresh
    },
    onSessionError: (err) => {
      // Handle session error
    },
    onSessionClosed: () => {
      // Handle session closure
    };
})
```

### DIFF
--- a/src/components/Chat.tsx
+++ b/src/components/Chat.tsx
@@ -41,7 +41,7 @@ const Chat = () => {
     stores.sdkStore.initialized,
   ]);
 
-  if (!widgetSession.channelUrl) return <></>;
+
 
   const onBeforeSendMessage = <
     T extends UserMessageCreateParams | FileMessageCreateParams
@@ -60,7 +60,7 @@ const Chat = () => {
 
   return (
     <GroupChannelProvider
-      channelUrl={widgetSession.channelUrl}
+      channelUrl={widgetSession.channelUrl ?? ''}
       scrollBehavior={'smooth'}
       onBeforeSendUserMessage={onBeforeSendMessage}
       onBeforeSendFileMessage={onBeforeSendMessage}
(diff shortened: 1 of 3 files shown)


src/components/Chat.tsx
```diff
@@ -60,7 +60,7 @@ const Chat = () => {
 
   return (
     <GroupChannelProvider
-      channelUrl={widgetSession.channelUrl}
+      channelUrl={widgetSession.channelUrl ?? ''}
```
bang9, line 63: ### Before
image

### After
image

src/context/ConstantContext.tsx
```diff
@@ -74,7 +105,7 @@ export const ConstantStateProvider = (
          * userId & configureSession should be used together to create a group channel on the client side.
          */
         userId: props.userId,
-        configureSession: props.configureSession,
+        configureSession: props.configureSession ? configureSession : undefined,
```
chohongm, line 108: hmm why can't it be just `configureSession: configureSession`?

AhyoungRyu: Cause we're using trying to use the custom one if props.configureSession is provided?

AhyoungRyu (approved):

src/context/ConstantContext.tsx
chohongm: if props.configureSession not given,  new SessionHandler() will be set to configureSession. So I thought  we can just inject configureSession to provider.

bang9: configureSession is optional. If it’s not provided, it shouldn't be included.

Below code is just to avoid type error, it is unrechable code
```ts
if (!handler) return new SessionHandler();
```

PR #247, opened 2024-05-27
[AC-2463] docs: update custom-session-guide.md

bang9:
- rendered version: https://github.com/sendbird/chat-ai-widget/blob/docs/configure-session/custom-session-guide.md
- ticket: [AC-2463]

[AC-2463]: https://sendbird.atlassian.net/browse/AC-2463?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/custom-session-guide.md
+++ b/custom-session-guide.md
@@ -18,7 +18,7 @@ const API_TOKEN = "YOUR_API_TOKEN";
 
 const issueSessionToken = async (
   userId: string,
-  period = 10 * 60 * 1000
+  expiryDuration = 10 * 60 * 1000
 ): Promise<string> => {
   const url = `https://api-${APP_ID}.sendbird.com/v3/users/${userId}/token`;
   const response = await fetch(url, {
@@ -27,7 +27,7 @@ const issueSessionToken = async (
       "Content-Type": "application/json; charset=utf8",
       "Api-Token": API_TOKEN,
     },
-    body: JSON.stringify({ expires_at: Date.now() + period }),
+    body: JSON.stringify({ expires_at: Date.now() + expiryDuration }),
   });
 
   const data = await response.json();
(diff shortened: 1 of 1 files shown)


custom-session-guide.md
```diff
@@ -1,67 +1,102 @@
...
+
+## Step-by-Step Guide
+
+### Defining a Function to Issue Session Tokens
+
+Define a function to issue session tokens. The following TypeScript code is an example; in practice, you should issue session tokens from your server. For more information, please see the [Sendbird documentation](https://sendbird.com/docs/chat/platform-api/v3/user/managing-session-tokens/issue-a-session-token).
+
+```ts
+const APP_ID = "YOUR_APP_ID";
+const API_TOKEN = "YOUR_API_TOKEN";
```
AhyoungRyu, line 17: Would it be better if we mention where the user can get this api token? https://sendbird.com/docs/chat/platform-api/v3/prepare-to-use-api#2-authentication

I remember I looked up the information in the official docs page when I had to do this

AhyoungRyu (approved):
Looks great!

custom-session-guide.md
bang9: Actually, I mention the link https://sendbird.com/docs/chat/platform-api/v3/prepare-to-use-api under **Prerequisites**

bang9:
Configure session sample link: https://github.com/sendbird/chat-ai-widget-session-sample/pull/1

custom-session-guide.md
```diff
@@ -1,67 +1,102 @@
...
+
+const issueSessionToken = async (
+  userId: string,
+  period = 10 * 60 * 1000
```
chohongm: minor: I think period is not the best choice of parameter name in this case because period concerns a cyclic event. However, in this case, this is used only one time for setting a expiry datetime. Maybe expireFromNow is better name?

chohongm (approved):
Nice guide! LGTM

custom-session-guide.md
bang9: how about `expiryDuration`?

PR #248, opened 2024-05-27
[AC-2476] fix: configure session

### LINKED ISSUE #243: Setting configureSession results in a broken Chatbot
I'm trying to use `configureSession` on `<ChatAiWidget />` but I can't get it to work.
I've followed the [step-by-step guide](https://github.com/sendbird/chat-ai-widget/blob/develop/custom-session-guide.md) in the documentation but when I set the `configureSession` prop on `<ChatAiWidget />`, the chatbot window gets empty and transparent (see screenshot below).
Capture d’écran 2024-05-23 à 10 20 55

No console logs are seen even though all session callbacks are issuing logs.

See component code below
```
export default function SendbirdChatbot({
  userEmail,
}: SendbirdChatbotProps): React.JSX.Element | null {
  const applicationId = import.meta.env.VITE_SENDBIRD_APPLICATION_ID as
    | string
    | undefined
  const botId = import.meta.env.VITE_SENDBIRD_BOT_ID as string | undefined
  const apiHost = import.meta.env.VITE_SENDBIRD_API_HOST as string | undefined

  const issueSessionToken = useCallback(async (userId: string): Promise<string> => {
    const url = `https://api-${applicationId}.sendbird.com/v3/users/${encodeURIComponent(userId)}/token`
    const response = await fetch(url, {
      method: 'POST',
      headers: {
        'Content-Type': 'application/json; charset=utf8',
        'Api-Token': '<API-TOKEN>'
      }
    })

    const data = await response.json()
    if (response.ok) {
      return data['token'] as string
    } else {
      throw new Error(`${data['code']} - ${data['message']}`)
    }
  }, [applicationId])

  const configureSession = useCallback(
    (): SessionHandler => {
      const sessionHandler = new SessionHandler()

      sessionHandler.onSessionTokenRequired = (resolve, reject) => {
        console.log('SessionHandler.onSessionTokenRequired()')
        issueSessionToken(userEmail ?? '').then(token => resolve(token)).catch(error => reject(error))
      }
      sessionHandler.onSessionRefreshed = () => {
        console.log('SessionHandler.onSessionRefreshed()')
      }
      sessionHandler.onSessionError = (error) => console.log(error)
      sessionHandler.onSessionClosed = () => console.log('SessionHandler.onSessionClosed()')

      return sessionHandler
    }, [issueSessionToken, userEmail]
  )

  return applicationId && botId && apiHost ? (
    <ChatAiWidget
      applicationId={applicationId}
      botId={botId}
      apiHost={apiHost}
      userId={userEmail}
      configureSession={configureSession}
    />
  ) : null
}
```

NOTE: this is only a POC, we are aware that we should not directly use the `Api-Token` client side. In the production version the session token will be issued from our backend.

bang9:
## Changes
- [UIKit] merge `main` into `ai-widget/experimental` branch and update HEAD.
- Do not reuse the cached session if the user ID has been changed in manual strategy.

ticket: [AC-2476]
resolve: #243

[AC-2476]: https://sendbird.atlassian.net/browse/AC-2476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/context/WidgetSettingContext.tsx
+++ b/src/context/WidgetSettingContext.tsx
@@ -85,10 +85,16 @@ export const WidgetSettingProvider = ({
       botId,
     });
 
-    const reuseCachedSession =
-      !!cachedSession &&
-      cachedSession.strategy === strategy &&
-      !isPastTime(cachedSession.expireAt);
+    const reuseCachedSession = ((cache: typeof cachedSession): cache is NonNullable<typeof cachedSession> => {
+      if (!cache || cache.strategy !== strategy) return false;
+      if (cache.strategy === 'manual') {
+        return cache.userId === injectedUserId;
+      }
+      if (cache.strategy === 'auto') {
+        return !isPastTime(cache.expireAt);
+      }
+      return false;
+    })(cachedSession);
 
     const response = await getWidgetSetting({
       host: apiHost,
--- a/src/context/WidgetSettingContext.tsx
+++ b/src/context/WidgetSettingContext.tsx
@@ -88,7 +88,9 @@ export const WidgetSettingProvider = ({
     const reuseCachedSession = ((cache: typeof cachedSession): cache is NonNullable<typeof cachedSession> => {
       if (!cache || cache.strategy !== strategy) return false;
       if (cache.strategy === 'manual') {
-        return cache.userId === injectedUserId;
+        // NOTE: There is no need to check the expiration of the session if it is managed manually.
+        // However, since the existing logic has been regenerating the channel every 30 days due to this logic.
+        return !isPastTime(cache.expireAt) && cache.userId === injectedUserId;
       }
       if (cache.strategy === 'auto') {
         return !isPastTime(cache.expireAt);


src/context/WidgetSettingContext.tsx
```diff
@@ -85,10 +85,16 @@ export const WidgetSettingProvider = ({
...
     });
 
-    const reuseCachedSession =
-      !!cachedSession &&
-      cachedSession.strategy === strategy &&
-      !isPastTime(cachedSession.expireAt);
+    const reuseCachedSession = ((cache: typeof cachedSession): cache is NonNullable<typeof cachedSession> => {
+      if (!cache || cache.strategy !== strategy) return false;
+      if (cache.strategy === 'manual') {
+        return cache.userId === injectedUserId;
```
bang9: We don’t check the user’s expiry here because it is handled by the customers.

bang9: However, it might make sense if we need to reset the channel every 30 days. 🤔

bang9: bring back the logic.

src/context/WidgetSettingContext.tsx
```diff
@@ -85,10 +85,18 @@ export const WidgetSettingProvider = ({
...
+    const reuseCachedSession = ((cache: typeof cachedSession): cache is NonNullable<typeof cachedSession> => {
+      if (!cache || cache.strategy !== strategy) return false;
+      if (cache.strategy === 'manual') {
+        // NOTE: There is no need to check the expiration of the session if it is managed manually.
+        // However, since the existing logic has been regenerating the channel every 30 days due to this logic.
+        return !isPastTime(cache.expireAt) && cache.userId === injectedUserId;
+      }
+      if (cache.strategy === 'auto') {
+        return !isPastTime(cache.expireAt);
+      }
```
AhyoungRyu, line 97:
```suggestion
      const isValidExpiration = !isPastTime(cache.expireAt);
      const isUserMatching = cache.userId === injectedUserId;
      if (cache.strategy === 'auto') return isValidExpiration;
      // NOTE: There is no need to check the expiration of the session if it is managed manually.
      // However, since the existing logic has been regenerating the channel every 30 days due to this logic.
      if (cache.strategy === 'manual') return isValidExpiration && isUserMatching;
```

nit; suggesting this since it seems you prefer a single line if statement ;)

AhyoungRyu (approved):

bang9:
Configure session sample link: https://github.com/sendbird/chat-ai-widget-session-sample/pull/1

PR #251, opened 2024-05-27
feat: Added messageInputControl to Constant

chohongm:
### Changelog
- Added `messageInputControlProps` to `Constant`

### Ticket
https://sendbird.atlassian.net/browse/AC-2498

### Discussion
https://sendbird.slack.com/archives/C0585965FFA/p1716798555294749?thread_ts=1716795864.100449&cid=C0585965FFA


src/hooks/useDisableInputUntilReply.ts
```diff
@@ -38,12 +47,7 @@ export const useDisableInputUntilReply = ({
         setIsMessageInputDisabled(false);
       }
     }
-    return () => {
-      if (timerRef.current) {
-        clearTimeout(timerRef.current);
-      }
-    };
-  }, [currentUserId, botUser, lastMessage]);
```
chohongm, line 46: I had to remove this because this is being called upon receiving the very first streaming bot message. Such that timer is cleared when it should not be. Feel free to test it out to see what I mean

src/const.ts
```diff
@@ -233,6 +243,10 @@ export interface Constant extends Required<ConstantProps> {
    * @private Callback to be called when the widget open state changes.
    */
   onWidgetOpenStateChange?: (params: OnWidgetOpenStateChangeParams) => void;
+  /**
+   * @private Message input state controlling properties.
+   */
+  messageInputControlProps?: MessageInputControlProps;
```
bang9: How about avoiding the use of the term "Props" in contexts where it refers to parameters of general functions or specific options, since in a React development environment, it specifically means attributes passed to a component?

Also, it feels a bit awkward to use "Props" again within a props.

src/const.ts
```diff
@@ -233,6 +243,10 @@ export interface Constant extends Required<ConstantProps> {
    * @private Callback to be called when the widget open state changes.
    */
   onWidgetOpenStateChange?: (params: OnWidgetOpenStateChangeParams) => void;
+  /**
+   * @private Message input state controlling properties.
```
bang9: Isn't this feature part of the public interface?

src/const.ts
```diff
@@ -120,6 +120,16 @@ export interface BotStudioEditProps {
   suggestedRepliesDirection?: 'horizontal' | 'vertical';
 }
 
+export interface MessageInputControlProps {
```
bang9: Explicit naming is good, but it would be better to minimize parts that can be quite clearly inferred implicitly. For instance, "message input" could be simplified to just "input"

src/const.ts
```diff
@@ -120,6 +120,16 @@ export interface BotStudioEditProps {
...
+  /**
+   * If undefined, false, or 0 is given, message input will not be blocked.
+   * If true or non-zero numeric value is given, message input will be blocked.
+   * If true is given, there will not be a force unblock.
+   * If numeric value is given, it will be used as timeout to force unblock the input.
+   */
+  blockInputWhileBotResponding?: boolean | number;
```
bang9: This is already an option under input control. Wouldn't it be okay to exclude "input"?
`blockWhileBotResponding`

bang9 (commented):
Additionally, it's best to avoid the repetitive use of keywords in nested structures as much as possible.

example
```js
{
  "buttonSettings": {
    "buttonColor": "blue",
    "buttonSize": "large",
    "buttonLabel": "Submit"
  }
}
```
vs
```js
{
  "buttonSettings": {
    "color": "blue",
    "size": "large",
    "label": "Submit"
  }
}
```

src/const.ts
chohongm: oh input could mean many things so I thought necessary to use messageinput here. But since many of you think otherwise, I will remove it!

src/const.ts
chohongm: ah I was wondering what was the meaning of this because I thought all Constant props are public. Ill change

src/const.ts
chohongm: I see. Do you have any suggestions? I thought of messageInputControlAttributes.

src/const.ts
AhyoungRyu: Mark it as public if you think a prop is okay to write in REAME and a real users are okay to use it. In this case, don't forget to add it to the prop table in README.
But a prop has been added just for an internal usage (i.e. bot studio), then mark it as private :)

src/hooks/useDisableInputUntilReply.ts
```diff
@@ -38,12 +47,7 @@ export const useDisableInputUntilReply = ({
...
+  }, [currentUserId, botUser, lastMessage, blockInputWhileBotResponding]);
```
AhyoungRyu:
```suggestion
  }, [currentUserId, botUser?.userId, lastMessage?.messageId, blockInputWhileBotResponding]);
```

Better to keep the items in string format since React uses shallow equality comparison to identify whether dependencies are updated. 

https://dmitripavlutin.com/how-to-compare-objects-in-javascript/#3-shallow-equality

src/const.ts
bang9: `messageInputControls` or `inputControls`?

Personally, I'm not sure if there will be additional options to control the input in the future, so I prefer to avoid nested objects.

chohongm: Yep we can try avoid nested object interface. I think `disabled` is likely to be requested in the future. 
```
MessageInputControls {
  // disabled?: boolean; could be added in the future.
  blockWhileBotResponding?: boolean | number;
}
```

src/hooks/useDisableInputUntilReply.ts
chohongm: in this case, `lastMessage` is intended becaue of streaming. messageId does not change nor updatedTs. It is message and streaming state

src/const.ts
bang9: Sure, that sounds make sense. Additionally, we would like to receive the placeholder text for each case where it will be disabled in the future.

src/hooks/useDisableInputUntilReply.ts
```diff
@@ -8,6 +8,7 @@ interface UseDisableInputUntilReplyProps {
   lastMessage?: SendableMessage;
   botUser?: Member;
   currentUserId: string | null;
+  blockInputWhileBotResponding?: boolean | number;
```
AhyoungRyu: Q. This prop doesn't sound it can have a number value. Which case can it be a number?

README.md
```diff
@@ -108,18 +108,19 @@ export default App;
...
+| `configureSession`      | `() => SessionHandler`                                                                            | No      | N/A           | Function to configure the user session. The example usage can be found [here](./custom-session-guide.md). <br/><br/> Must be used with `userId`.          |
+| `autoOpen`              | `boolean`                                                                                         | No      | N/A        | Determines whether the chatbot widget automatically opens when the browser window is opened.                                                              |
+| `messageInputControls`  | `{ blockWhileBotResponding?: boolean \| number }`                                                 | No      | N/A        | Allows to control enabled/disabled state of the message input.                                                                                            |
```
bang9: how about

```markdown
| `messageInputControls.blockWhileBotResponding` | ~~ | ~~ |
```

bang9: For components, it might be acceptable to represent them as objects
but for options, it seems more convenient to separate each item individually.

src/utils/index.ts
```diff
@@ -332,6 +332,7 @@ export async function openURL(url?: string | null) {
 }
 
 export function isDashboardPreview(userAgent: object | undefined) {
+  return true;
```
bang9:
```suggestion
```

bang9: Always make sure to check for any unnecessary code before requesting review.

Before requesting a review, please make sure to check the following items:
- Ensure that the CI processes are all running correctly.
- Check that there are no unnecessary elements included in the code.

src/hooks/useDisableInputUntilReply.ts
```diff
@@ -17,18 +18,26 @@ export const useDisableInputUntilReply = ({
...
+      !botUser ||
+      !lastMessage
+    )
+      return;
     if (isSentBy(lastMessage, currentUserId) && lastMessage.messageId === 0) {
       setIsMessageInputDisabled(true);
-      timerRef.current = setTimeout(() => {
-        setIsMessageInputDisabled(false);
-      }, 10000);
+      if (typeof blockInputWhileBotResponding === 'number') {
```
bang9: If it is used as true, what is the default duration in seconds?

bang9: Also, it seems there is no logic that executes when it is used as true.

src/hooks/useDisableInputUntilReply.ts
chohongm: its used as timeout value to force unblock blocked message input in case reply is never received
https://sendbird.slack.com/archives/C0585965FFA/p1716798555294749?thread_ts=1716795864.100449&cid=C0585965FFA

src/utils/index.ts
chohongm: oh my mistake. Thanks for the catch! Ill be careful of this.

src/hooks/useDisableInputUntilReply.ts
chohongm: I added note to the exported interface
```
export interface MessageInputControls {
  /**
   * If undefined, false, or 0 is given, message input will not be blocked.
   * If true or non-zero numeric value is given, message input will be blocked.
   * If true is given, there will not be a force unblock.
   * If numeric value is given, it will be used as timeout to force unblock the input.
   */
  blockWhileBotResponding?: boolean | number;
}
```
> If true is given, there will not be a force unblock.

I wanted to give them option to turn off force unblock

bang9: Oh, I didn't see that. Thank you for letting me know.

src/hooks/useDisableInputUntilReply.ts
bang9, line 1: hook 이름을 변경된 옵션 `blockWhileBotResponding` 에 맞추는건 어떨까요?

src/hooks/useDisableInputUntilReply.ts
bang9: currentUserId 랑 blockInputWhileBotResponding 같은 경우 파라메터로 받지 않고
훅 안에서 context 접근해서 바로 쓸 수 있을것 같습니다.

src/hooks/useDisableInputUntilReply.ts
```diff
@@ -17,18 +18,26 @@ export const useDisableInputUntilReply = ({
...
   useEffect(() => {
-    if (!currentUserId || !botUser || !lastMessage) return;
+    if (
+      !blockInputWhileBotResponding ||
+      !currentUserId ||
+      !botUser ||
+      !lastMessage
+    )
+      return;
     if (isSentBy(lastMessage, currentUserId) && lastMessage.messageId === 0) {
```
bang9: 여기 로직에, sending status 가 failed 된 경우 클리어 하는 로직을 추가해주실 수 있을까요?

chohongm: 오 넵 좋은 제안 감사합니다!

chohongm: 휴 수정 반영했고 테스트가 조금 복잡한데 잘 동작합니다

https://github.com/sendbird/chat-ai-widget/assets/16806397/4a890df0-de7c-4498-b4de-d682fc0e7221

bang9: 고생하셨습니다!

추가로 생각을 해보았는데 failed/pending 메세지들은 항상 마지막에 나타나야 하기 때문에, lastMessage 에 항상 failed/pending 메세지가 와서

failed 가 하나 생긴 이후에는 block 동작이 잘 안될 것 같다는 생각이.. 드네요 🥲
일단 known 이슈로 두고 넘어가시지요..!

README.md
```diff
@@ -108,18 +108,19 @@ export default App;
...
+| `customRefreshComponent` | `{ icon: string, style: React.CSSProperties, width: string, height: string, onClick: () => void }` | No      | N/A           | Customizable refresh component. You can set properties such as `icon`, `style`, `width`, `height`, and `onClick`.                                                                       |
+| `configureSession`       | `() => SessionHandler`                                                                          | No      | N/A           | Function to configure the user session. The example usage can be found [here](./custom-session-guide.md). <br/><br/> Must be used with `userId`.                                        |
+| `autoOpen`               | `boolean`                                                                                       | No      | N/A        | Determines whether the chatbot widget automatically opens when the browser window is opened.                                                                                            |
```
bang9: nit: 여기 다른 테이블쪽의 변경은 되돌려주실 수 있을까요? 다른 기능들 머지할때 컨플릭트가 생길 것 같아요

bang9 (approved):
lgtm!

src/hooks/useDisableInputUntilReply.ts
chohongm: 오 넵 messages, localMessages 로 나뉘면 해결할수 있을것 같습니다. tools 에서 수정해야될것 같네요

README.md
chohongm: 요게 밑에 추가할때 max string length 에 따라서 자동으로 전체가 맞춰지더라구요 ('|'요거가 일자로 align 되야하는것 같음).

bang9: IDE 옵션일거에유

chohongm: done!

PR #255, opened 2024-05-30
[AC-2433] feat: hide widget for deactivated user

bang9:
## Feature

https://github.com/sendbird/chat-ai-widget/assets/26326015/acc85f33-a33a-4a8a-b238-b81a647e3ab1

Added the `enableHideWidgetForDeactivatedUser` option.
The widget will not be displayed if a deactivated error occurs at the time of connection.

## Changes
- `applicationId` and `botId` are now required.
- `WidgetOpenContext` has been changed to `WidgetStateContext`, and the widget visible state has been added.
- Modified the rendering control of `WidgetToggleButton` to avoid side effects during conditional rendering.
  - The autoOpen logic has been moved externally to operate regardless of whether `WidgetToggleButton` is rendered, as it is not directly related to the `WidgetToggleButton` component.

ticket: [AC-2433]

[AC-2433]: https://sendbird.atlassian.net/browse/AC-2433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/ProviderContainer.tsx
```diff
@@ -71,42 +78,60 @@ const SBComponent = ({ children }: { children: React.ReactElement }) => {
...
-        </ThemeProvider>
-      </StyleSheetManager>
-    </WidgetOpenProvider>
+              },
+            }}
+            uikitOptions={{
+              groupChannel: {
+                input: {
+                  // To hide the file upload icon from the message input
+                  enableDocument: false,
```
chohongm, line 123: nice!

src/const.ts
```diff
@@ -65,6 +65,7 @@ export const DEFAULT_CONSTANT = {
   enableEmojiFeedback: true,
   enableMention: true,
   enableResetHistoryOnConnect: false,
+  enableHideWidgetForDeactivatedUser: false,
```
chohongm, line 71: if user is deactivated and enableHideWidgetForDeactivatedUser is false, what happens when widget is open? (can channel be fetched? is message input disabled? etc)

src/const.ts
```diff
@@ -235,11 +224,27 @@ export interface Constant extends Required<ConstantProps> {
   onWidgetOpenStateChange?: (params: OnWidgetOpenStateChangeParams) => void;
 }
 
-export interface ConstantProps {
+interface ConstantFeatureFlags {
```
chohongm, line 248: nice

src/context/ConstantContext.tsx
```diff
@@ -8,19 +8,21 @@ import {
...
 
 import { type Constant, DEFAULT_CONSTANT, widgetStringSet } from '../const';
 import { getDefaultServiceName, isMobile } from '../utils';
 
 const initialState = DEFAULT_CONSTANT;
 
-interface ConstantContextProps extends Partial<Constant> {
-  applicationId: string | null;
-  botId: string | null;
+interface ConstantContextProps extends Omit<Partial<Constant>, 'stringSet'> {
```
chohongm, line 18: what is goal of omit stringSet?

src/hooks/useWidgetAutoOpen.ts
```diff
@@ -0,0 +1,27 @@
+import { useEffect, useRef } from 'react';
+
+import { useConstantState } from '../context/ConstantContext';
+import { useWidgetState } from '../context/WidgetStateContext';
+
+export function useWidgetAutoOpen() {
```
chohongm, line 6: nice

src/main.tsx
```diff
@@ -2,10 +2,12 @@ import React from 'react';
...
 import App from './App';
-import { DEFAULT_CONSTANT } from './const';
 
 ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
   <React.StrictMode>
-    <App firstMessageData={DEFAULT_CONSTANT.firstMessageData} />
+    <App
+      applicationId={import.meta.env.VITE_CHAT_WIDGET_APP_ID}
+      botId={import.meta.env.VITE_CHAT_WIDGET_BOT_ID}
+    />
```
chohongm, line 11: nice

src/const.ts
bang9: connection fails and nothing will work
image

src/context/ConstantContext.tsx
bang9: to avoid type error

src/const.ts
chohongm: hmm not sure when client would want this. Hence maybe enableHideWidgetForDeactivatedUser is not necessary. just set enableHideWidgetForDeactivatedUser as true internally and do not open this to public. What do you think?

src/context/ConstantContext.tsx
chohongm: no i meant since stringSet is already in Constant why omit here and add to ConstantContextProps?

bang9: `ConstantContextProps` type should takes the stringSet as `Partial<StringSet>` type.
However, since `StringSet` is declared in the `Constant` type, it should be omitted, and the `Partial<StringSet>` type should be added to avoid type errors.

src/const.ts
bang9: This feature determines the button's visibility after the connection is made, resulting in a slight delay equal to the connection time in rendering the button.
Additionally, since this delay will affect customers who do not use the deactivate user feature, it should not be enabled by default.

chohongm: ok but I think just displaying "somethign went wrong retry button" in case of deactivated user is misleading. We have to do something about this.

src/context/ConstantContext.tsx
chohongm: ah Partial => optional. got it, thanks!

chohongm (approved):
LGTM. Very nice PR description

src/const.ts
bang9: Usually, it is rare to deactivate specific users while using the widget.
If there is a request to hide the widget from malicious users, instructions on how to use this feature and deactivate users will be provided.

PR #261, opened 2024-06-03
[AC-2572] feat: widget_setting by strategy

### LINKED ISSUE #266: configureSession.onSessionTokenRequired is not triggered
We are having issues getting `onSessionTokenRequired` to be called. We were able to replicate this issue using the real world example found in https://github.com/sendbird/chat-ai-widget-session-sample

The problem is that `onSessionTokenRequired` is only called if the `userId` was automatically created by the Sendbird chatbot.
This can be achieved by doing the following:
* in the sample app, remove or comment the `userId` and `configureSession` prop from the `ChatAiWidget` component
* run the sample app
* take note of the automatically assigned user id (this can be seen in the Network tab of the browser's Developer Tools). It uses the following format: `widget_<random 32 hexadecimal character string>`
* in the sample app, change the `USER_ID` variable to the previously obtained user id and uncomment the `userId` and `configureSession` props
* run the sample app again
* notice in the Console logs that `onSessionTokenRequired` is being called

Now if you change `USER_ID` to be any other user id and run the sample app, you won't see `onSessionTokenRequired` being called in the Console logs. We've tried the following cases:
* create a user using Sendbird's REST API and then set `USER_ID` to the newly created user's id in the sample app
* set `USER_ID` in the sample app to a brand new id that we have come up with. In this case a user with this id gets created in Sendbird, but we still don't see `onSessionTokenRequired`. being triggered.

Is this expected ?

bang9:
## Description

| Cache State | strategy   | Request Parameters                                | Response                                        |
|-------------|--------|---------------------------------------------------|----------------------------------------------------|
| Cache X     | auto   | `?create_user_and_channel=True`                   | - bot style<br>- Creates and returns a new user.<br>- Creates and returns a new channel.  |
| Cache O     | auto   | `?user_id={already_created_user_id}`              | - bot style<br>- Creates and returns a new channel if needed.                            |
| Cache X     | manual | `?create_channel=True&user_id={custom_user_id}`   | - bot style<br>- Creates and returns a new channel.        |
| Cache O     | manual   | `?user_id={custom_user_id}`              | - bot style<br>- Creates and returns a new channel if needed.                            |

- [x] Update `manual + cache o` case

ticket: [AC-2572]

## Changes
- Added `sessionToken` to widget props for configure custom session
- Added session flow to support channel update

## Note
- `Manual + Cache O`, `Manual + Cache X` will be updated later. not implemented yet.
- resolve #266

[AC-2572]: https://sendbird.atlassian.net/browse/AC-2572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.eslintrc.cjs
```diff
@@ -50,6 +50,15 @@ module.exports = {
         'includeRoles': ['button']
       },
     ],
+    '@typescript-eslint/no-unused-vars': [
+      1,
+      {
+        vars: 'all',
+        varsIgnorePattern: '^_',
+        args: 'after-used',
+        argsIgnorePattern: '^_',
```
chohongm, line 59: we might want
eqeqeq: To enforce the use of === and !== over == and !=.

abstract equality operators can lead to human error
0 == '0' is true but 0 === '0' is false. ignores type in this case which can lead to logical error.

src/context/WidgetSettingContext.tsx
```diff
@@ -8,18 +8,19 @@ import React, {
...
 } from '../libs/storage/widgetSessionCache';
-import { getDateNDaysLater, isPastTime } from '../utils';
+import { getDateNDaysLater, isPastTime, noop } from '../utils';
 
 interface WidgetSession {
   strategy: 'auto' | 'manual';
   userId: string;
   expireAt: number;
-  // channelUrl is optional and is dynamically generated for a manual strategy.
+  // channelUrl is optional and is dynamically generated for a legacy manual strategy.
```
chohongm, line 23: why has this become legacy? and what is new manual strategy? is there reference we can read?

update: Please ignore this comment. checking below logic

.eslintrc.cjs
bang9: I did not understand. This is a rule with some additional settings for the no-unused-vars rule.

```ts
function exampleFunction(_unusedArg: string, usedArg: number) {
  const _unusedVar = 'I am not used';
  console.log(usedArg);
}
```
In this example, the `_unusedArg` and `_unusedVar` are prefixed with an underscore, so they are ignored by the no-unused-vars rule as specified in the configuration.

bang9: Are you asking to add a lint rule for using `===`?

chohongm: oh i wanted to hear your thoughts on adding a lint rule for using ===. Personally, I find using both === and == confusing and personally I find not using == is fine.

src/context/WidgetSettingContext.tsx
```diff
@@ -82,7 +83,7 @@ export const WidgetSettingProvider = ({
       botId,
     });
 
-    const reuseCachedSession = ((
+    const useCachedSession = ((
```
chohongm, line 91: Question on:

// However, since the existing logic has been regenerating the channel every 30 days due to this logic.

Is this for backward compat?

src/context/WidgetSettingContext.tsx
```diff
@@ -97,49 +98,83 @@ export const WidgetSettingProvider = ({
...
+      //   if (handler.onSessionTokenRequired) {
+      //     return (
+      //       (await new Promise(handler.onSessionTokenRequired)) ?? undefined
+      //     );
+      //   }
+      // } catch {
+      //   // noop
+      // }
+      // return undefined;
+    })();
```
chohongm: so for manual, bot styles will always be fetched from now on?

.eslintrc.cjs
bang9: right. I completely agree with the use of `===`. I will add the rule (separate PR).

chohongm: if we are alreadying using == in the project and not so simple to replace it, we can take this separate task

src/context/WidgetSettingContext.tsx
bang9: Rather than backward compat, we could remove the logic that checks the expiration date.
However, since a new channel is created after a fairly long period of 30 days, I believe it is quite reasonable to keep it as is.

src/context/WidgetSettingContext.tsx
bang9: Before and after, the `botStyle` always be fetched.

src/libs/api/widgetSetting.ts
```diff
@@ -92,3 +102,145 @@ export async function getWidgetSetting({
...
+        // Legacy manual
+        callbacks.onManualNonCached();
+      }
+    }
+  }
+
+  return handlers;
+};
+
+function getParamsByStrategy(
```
chohongm, line 221: minor: What do you think about adding comment to the function?
```
/**
 * When manual strategy, regardless of cached session, userId must be injected and channel will be
 * created "manually" through sdk once sdk is ready (IFF channelUrl is not cached as session). See initManualSession() for reference.
 */
```

bang9: The logic for creating channels in the SDK will likely be removed. In the future, the logic for managing sessions will operate based on the shared session flow chart. Since this includes internal(server) policies, it might be better to share it through internal documents.

And the mentioned content is noted here: https://github.com/sendbird/chat-ai-widget/pull/261/files#diff-362487901b9c6302b2d7c52add72f6440f65334a5261fd09d24ce36fb4a7970dR162-R165

src/context/WidgetSettingContext.tsx
chohongm: hmm unless there is a client that come back to channel after 1 year and see the history. There is a way for this? then its ok I think. I think this is very unlikely and we can ignore

src/context/WidgetSettingContext.tsx
chohongm: ah before was also returning undefined. the old logic is hard to grasp. Might as well just remove it and change to
```
 await widgetSettingHandler(strategy, useCachedSession, {
      host: apiHost,
      appId,
      botId,
      userId: strategy === 'manual' ? injectedUserId : cachedSession?.userId,
      sessionKey: undefined,
    })
```

src/libs/api/widgetSetting.ts
chohongm: ok good!

chohongm (approved):
LGTM. Nicely done!

src/context/WidgetSettingContext.tsx
bang9: Since this seems related to the widget policy, let me ask for opinions on the channel!

bang9: Note: keep 30 days

bang9:
I'll merge this after server deployment.

bang9:
The policy of the manual strategy on the table will be applied later.

src/libs/api/widgetSetting.ts
```diff
@@ -231,7 +231,8 @@ function getParamsByStrategy(
     }
   } else {
     if (useCachedSession) {
-      return { userId: params.userId };
+      return {};
+      // return { userId: params.userId };
```
chohongm, line 235: is this because manual reuses channel only?

src/context/WidgetSettingContext.tsx
```diff
@@ -159,7 +159,7 @@ export const WidgetSettingProvider = ({
             ...cachedSession,
             channelUrl: channel?.channelUrl ?? cachedSession.channelUrl,
           } satisfies WidgetSessionCache;
-          setWidgetSession(session);
+          setWidgetSession({ ...session, sessionToken });
```
chohongm, line 162: what about `onAutoCached`. No need to save sessionToken there?

src/libs/api/widgetSetting.ts
bang9: image

src/context/WidgetSettingContext.tsx
bang9: You can check detailed information from here: https://sendbird.atlassian.net/wiki/spaces/SDK/pages/2391277608/Widget+Session+management+strategy

chohongm: ah got it now, thanks. for auto, sessionToken is created in getWidgetSetting and cached in callback

chohongm (approved):
LGTM!

PR #263, opened 2024-06-04
[AC-2264] fix: handle channel not found error case

bang9:
## Changes
Reset the session when the channel is not found.

https://github.com/sendbird/chat-ai-widget/assets/26326015/8c0c6432-2546-4dfc-aa58-20e2605f55dc

ticket: [AC-2264]

[AC-2264]: https://sendbird.atlassian.net/browse/AC-2264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/context/WidgetSettingContext.tsx
```diff
@@ -195,6 +200,7 @@ export const WidgetSettingProvider = ({
         },
         widgetSession,
         initManualSession,
+        resetSession: () => initSessionByStrategy(sessionStrategy, true),
```
chohongm, line 203: Shouldn't we be handling error here? onChannelFetchFailed is called by UIKit but initSessionByStrategy does not fetch channel through UIKit so failing here would not automatically call onChannelFetchFailed

bang9: ~~To do so, we need to make an unnecessary API call just to check if the channel exists.~~

initSessionByStrategy is not an API to get channels; it's a role to create them.

chohongm: ok yes but my point is not about GET or POST but about synchronizing the retries. Maybe I misunderstood but I thought

If uikit fails fetching channel, onChannelFetchFailed() will be called which then calls resetSession(). This tries to create channel. What happens if this fails?

chohongm: seeing resetReqCounter, I expected onChannelFetchFailed would be called over and over again upto five times, is resetSession() failure going to call onChannelFetchFailed ? bc I could not find connection between the two

bang9: > ok yes but my point is not about GET or POST but about synchronizing the retries. Maybe I misunderstood but I thought
> 
> If uikit fails fetching channel, onChannelFetchFailed() will be called which then calls resetSession(). This tries to create channel. What happens if this fails?

If it fails, nothing happens.

bang9: > seeing resetReqCounter, I expected onChannelFetchFailed would be called over and over again upto five times, is resetSession() failure going to call onChannelFetchFailed ? bc I could not find connection between the two

`resetSession` is a method that creates a new channel, and `onChannelFetchFailed` is a method that gets called when fetching the channel fails.

There is no direct connection between `resetSession` and `onChannelFetchFailed`, but the logic flow is cyclic.
>fetch channel -> not found channel -> `onChannelFetchFailed` -> `resetSession` -> channel created -> fetch channel -> good !

If the channel gets deleted in the middle of this flow, an infinite loop can occur.
>fetch channel -> not found channel -> `onChannelFetchFailed` -> `resetSession` -> channel created -> **channel deleted** -> fetch channel -> not found channel -> `onChannelFetchFailed` -> ... loop

I believe the probability of this happening in actual business logic is close to zero.
Nevertheless, since this logic has the potential to cause an infinite loop, this is a defensive code to prevent it.

src/components/CustomChannelComponent.tsx
```diff
@@ -238,9 +242,16 @@ export function CustomChannelComponent() {
...
+
   return (
     <Root height={'100%'} isStaticReplyVisible={isStaticReplyVisible}>
-      <ChannelUI
+      <GroupChannelUI
+        onChannelFetchFailed={() => {
+          if (resetReqCounter.current > 5) return;
+          resetReqCounter.current += 1;
+          resetSession();
+        }}
```
chohongm, line 254: I want to understand the case where onChannelFetchFailed is called multiple times such that this code is executed multiple times. Could you explain the case?

chohongm (approved):
LGTM!

src/components/CustomChannelComponent.tsx
bang9: >fetch channel -> not found channel -> `onChannelFetchFailed` -> `resetSession` -> channel created -> **channel deleted** -> fetch channel -> not found channel -> `onChannelFetchFailed` -> ... loop

PR #280, opened 2024-06-18
chore: convert null response to undefined

bang9:
## Changes
- `toggle_button_url` API response can be `null`, so I set it to undefined when using it


chohongm (approved):
LGTM!

PR #283, opened 2024-06-20
[AC-2708] fix: hide scroll bar

bang9:
## Changes
- Hide scroll bar

ticket: [AC-2708]

[AC-2708]: https://sendbird.atlassian.net/browse/AC-2708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/css/index.css
+++ b/src/css/index.css
@@ -49,10 +49,6 @@
   padding: 0 16px !important;
 }
 
-.sendbird-message-input .sendbird-message-input-text-field {
-  overflow-y: hidden;
-}
-
 .sendbird-modal__content {
   width: calc(100% - 20px);
 }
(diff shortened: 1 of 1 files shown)


src/css/index.css
```diff
@@ -49,10 +49,6 @@
   padding: 0 16px !important;
 }
 
-.sendbird-message-input .sendbird-message-input-text-field {
-  overflow-y: hidden;
```
chohongm, line 53: this was not working?

chohongm (approved):
LGTM!

src/css/index.css
bang9: Yes, it doesn't work according to the CSS specificity.
Moreover, this option is for hiding overflowed content. If you style it this way, scrolling becomes impossible.

PR #301, opened 2024-07-08
feat: widget modularization - label

bang9:

## Changes

- created the Label component.
- applied [Linaria](https://linaria.dev/), a runtime zero CSS-in-JS library. (it is for experimental purposes, so it may be removed in the future.)

ticket: [AC-2914]

[AC-2914]: https://sendbird.atlassian.net/browse/AC-2914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/foundation/colors/theme.ts
```diff
@@ -0,0 +1,161 @@
+// * Theme
```
chohongm, line 1: Any plan on using these?

bang9: Maybe in the very distant future...? Currently, it is simply commented out to check the key values.

vite.config.pages.ts
```diff
@@ -1,11 +1,19 @@
...
 
 export default defineConfig({
   plugins: [
     react(),
+    wyw({
+      include: ['**/*.{ts,tsx}'],
+      babelOptions: {
+        presets: ['@babel/preset-typescript', '@babel/preset-react', '@wyw-in-js/babel-preset'],
+      },
+    }),
```
chohongm, line 16: could you explain why we need this please?

wyw-in-js.config.cjs
```diff
@@ -0,0 +1,30 @@
...
+    //   if (duplicates[className]) {
+    //     // print error message in shell with bold text..
+    //     console.log(`\n\x1b[1m\x1b[33m[ERROR] Duplicate className: ${className} for ${title}\x1b[0m\n`);
+    //     throw new Error(`Duplicate className: ${className}`);
+    //   } else {
+    //     duplicates[className] = true;
+    //   }
+    // }
+    //
+    // return className;
```
chohongm, line 28: could you explain why we need this please?

src/foundation/Label.tsx
```diff
@@ -0,0 +1,255 @@
...
+// `;
+
+const styles = {
+  base: css`
+    font-stretch: normal;
+    font-style: normal;
+    letter-spacing: normal;
+    font-family: var(--sendbird-font-family-default);
+  `,
+  lg: css`
```
chohongm, line 89: nice library

vite.config.pages.ts
bang9: It is simply the configuration of the library. 
https://github.com/callstack/linaria/blob/master/docs/BUNDLERS_INTEGRATION.md#Vite

To understand how it works, I recommend studying the concept of Babel.

wyw-in-js.config.cjs
bang9: It is simply the configuration of the library. 

https://wyw-in-js.dev/configuration

vite.config.pages.ts
chohongm: ok but why do we need wyw-in-js library?

bang9: Currently, we have installed and applied the Linaria library as an experiment (a zero-runtime CSS-in-JS library). 
One of the configurations needed in the Linaria library is `wyw-in-js`.

The role of `wyw-in-js` includes general tools necessary for developing CSS-in-JS libraries (such as JavaScript parsers for converting JS to CSS, and various plugins for different bundlers).

Since Linaria is developed using this `wyw-in-js` library, other configurations are also done through `wyw-in-js`, so it needs to be installed together.

image

chohongm: Thanks for the explaination 👍

chohongm (approved):
LGTM

PR #308, opened 2024-07-18
[AC-2915] feat: widget modularization - placeholder

bang9:

## Changes

- Added Placeholder components.
  - `loading`, `error`, `no channels` and `no messages` types are added.
  - `no results`, `search in` and `searching` types are not added (not being used in widget)
- Split themed color vars
- Use `hash` name in self-service build instead of `file` name

ticket: [AC-2915]

[AC-2915]: https://sendbird.atlassian.net/browse/AC-2915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/foundation/colors/css.ts
```diff
@@ -1,151 +1,112 @@
...
+  oncontent2: 'var(--sb-content-2)',
+  oncontent_inverse1: 'var(--sb-content-inverse-1)',
+  oncontent_inverse2: 'var(--sb-content-inverse-2)',
+  primary: 'var(--sb-primary)',
+  error: 'var(--sb-error)',
+  secondary: 'var(--sb-secondary)',
+};
+export const themedColorVars = css`
+  .sendbird-theme--light & {
+    --sb-bg-1: var(--sendbird-light-onlight-01);
```
chohongm, line 18: is `--sb-bg-1` good name for `--sendbird-light-onlight-01`. because i bg stands for background but i saw many cases `--sendbird-light-onlight-01` were being used in non background colors such as border color and text color

chohongm: can we just use `--sb-light-1` ?

src/foundation/components/Loader.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+type Props = {
+  size?: string | number;
+};
+
+export const Loader = (props: SBUFoundationProps<Props>) => {
+  const { className, children, size = 26 } = props;
+  const localProps = useLocalProps(props);
+  return (
+    <Container className={cx(className, 'sendbird-loader')} size={size} {...localProps}>
```
chohongm, line 35: is it missing themedColorVars? just checking.

src/foundation/components/Placeholder/Placeholder.error.tsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+const buttonContainer = css`
+  all: unset;
+  display: flex;
+  cursor: pointer;
+  gap: 4px;
+`;
+
+export default PlaceholderError;
```
chohongm, line 39: is it missing themedColorVars? just checking.

src/foundation/components/Placeholder/index.tsx
```diff
@@ -0,0 +1,176 @@
...
+//               color={LabelColors.ONBACKGROUND_2}
+//             >
+//               {stringSet.NO_SEARCHED_MESSAGE}
+//             </Label>
+//           )}
+//         </div>
+//       )}
+//     </div>
+//   );
+// }
```
chohongm, line 176: Should we keep this comment?

chohongm (approved):
LGTM please check my comments before merging

src/foundation/components/Placeholder/Placeholder.error.tsx
bang9: The class name for themedColorVars is not used here.

src/foundation/components/Loader.tsx
bang9: The class name for themedColorVars is not used here.

src/foundation/components/Placeholder/index.tsx
bang9: I've left only the code for the part that needs to be implemented additionally.

src/foundation/colors/css.ts
bang9: At least in UIKit, there are no such cases. Currently, our primary objective is to directly port the code from UIKit, so we will maintain it as is. If such cases are added in the future, it will not be too late to consider them then.

Semantically, there is a slight difference between on-bg and bg, so I will change it to on-bg.

PR #310, opened 2024-07-22
[AC-2727] feat: widget modularization - scroll to bottom button

bang9:

## Changes

- Added ScrollToBottomButton component.
- Update structure of foundation components

ticket: [AC-2727]

[AC-2727]: https://sendbird.atlassian.net/browse/AC-2727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/foundation/components/Label/css.ts
```diff
@@ -0,0 +1,88 @@
...
+//     ${styles}
+//   }
+// `;
+//
+// const sendbirdLabelMobile = css`
+//   ${mobileMixin(`
+//     -webkit-user-select: none;
+//     -webkit-touch-callout: none;
+//   `)};
+// `;
```
chohongm, line 14: Should we keep these comments?

src/foundation/components/Label/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+      | 'subtitle1'
+      | 'subtitle2'
+      | 'body1'
+      | 'body2'
+      | 'button1'
+      | 'button2'
+      | 'button3'
+      | 'caption1'
+      | 'caption2'
+      | 'caption3';
```
chohongm, line 33: will these be resused? maybe define `ElementType`?

src/foundation/components/Label/index.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+type AsProp<C extends ElementType> = {
+  as?: C;
+};
+type PropsToOmit<C extends ElementType, P> = keyof (P & AsProp<C>);
+type PolymorphicComponentProps<C extends ElementType, Props = NonNullable<unknown>> = Props &
+  AsProp<C> &
+  Omit<React.ComponentPropsWithoutRef<C>, PropsToOmit<C, Props>>;
+
+type HTMLProps<C extends ElementType> = PolymorphicComponentProps<C, HTMLAttributes<HTMLElement>>;
```
chohongm, line 17: here and above is a bit complex to understand. Let me ask you later.

chohongm (approved):
LGTM

src/foundation/components/Label/css.ts
bang9: I've left only the code for the part that needs to be implemented additionally.

src/foundation/components/Label/index.tsx
bang9: This is simply for accurate type inference. To fully understand it, you need to study the inference in TypeScript further. If you do not wish to deep dive into it, you can just understand it as a part for type inference.

src/foundation/components/Label/index.tsx
bang9: I didn't quite understand what you are asking.

PR #315, opened 2024-07-26
fix: handle error in asSafeURL

### LINKED ISSUE #314: URL crashes the AI Chatbot and the whole page
Since we've upgraded to version `1.7.3` of the SDK, we're running into an issue that we didn't experience with version `1.6.1`

We have a function call that returns a URL, and this URL is included by the chatbot into its response to the user as follows (this is only an excerpt from the full message showing the Markdown-formatted URL)
```
Your renewal lease has been fully executed. You can find a copy of the renewal lease [here](https://res.cloudinary.com/darwin-homes/image/authenticated/v1/sandbox/Ewmtbp/7781%20Kbujv%20Il/0021-55-88/xlsbsxx-eqpcuwu.pdf?__cld_token__=exp=1721913788~hmac=2620cdef709694691711b33a68f7cf41465d12f8343c54f58a09167f479ea2e1).
```

As soon as the chatbot displays this part of the message, it crashes and the whole page with it (whole page gets blank).

The exception that can be seen in the console is:
```
Uncaught (in promise) URIError: URI malformed
    at decodeURIComponent (<anonymous>)
    at D0 (chunk-EPLYPIMA.js?v=63f1ceef:27766:11)
    at replacer (chunk-EPLYPIMA.js?v=63f1ceef:28817:538)
    at chunk-EPLYPIMA.js?v=63f1ceef:28806:59
    at Array.forEach (<anonymous>)
    at chunk-EPLYPIMA.js?v=63f1ceef:28805:20
    at chunk-EPLYPIMA.js?v=63f1ceef:28809:11
    at Array.forEach (<anonymous>)
    at chunk-EPLYPIMA.js?v=63f1ceef:28797:7
    at Array.forEach (<anonymous>)
```

bang9:
resolve #314

### DIFF
--- a/src/__tests__/utils/index.test.ts
+++ b/src/__tests__/utils/index.test.ts
@@ -4,6 +4,9 @@ describe('asSafeURL', () => {
   test('should return the same URL if it is already safe', () => {
     expect(asSafeURL('http://example.com')).toBe('http://example.com');
     expect(asSafeURL('https://example.com')).toBe('https://example.com');
+
+    expect(asSafeURL('https://example.com/path/path2')).toBe('https://example.com/path/path2');
+    expect(asSafeURL('https://example.com?test=%')).toBe('https://example.com?test=%');
   });
 
   test('should return a safe URL if it is not safe', () => {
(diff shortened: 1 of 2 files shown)


chohongm (approved):
LGTM

PR #312, opened 2024-07-22
feat: widget modularization - replace group channel module

bang9:
- message list
  - [x] loading placeholder
  - [x] empty placeholder
  - [x] error placeholder
    - [ ] ~~implement retry action?~~
  - chat ui
    - [x] scroll to bottom
    - [x] date separator
    - [x] message
    - [x] message input
    - [x] typing indicator bubble
    - [x] feedback
    - [x] suggested replies
      - [ ] ~~static version for docs bot~~
    - [ ] ~~modal~~
    - [x] bot studio props
      - [x] welcome messages

tickets: [AC-2917], [AC-2918], [AC-2919], [AC-2920]

[AC-2918]: https://sendbird.atlassian.net/browse/AC-2918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AC-2919]: https://sendbird.atlassian.net/browse/AC-2919?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AC-2920]: https://sendbird.atlassian.net/browse/AC-2920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AC-2917]: https://sendbird.atlassian.net/browse/AC-2917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #318, opened 2024-07-29
fix: Fix typing indicator bubble to depend on bot user typing status update

chohongm:
### Ticket
https://sendbird.atlassian.net/browse/AC-3394

### DIFF
--- a/src/components/CustomChannelComponent.tsx
+++ b/src/components/CustomChannelComponent.tsx
@@ -1,7 +1,7 @@
 import { SendableMessage } from '@sendbird/chat/lib/__definition';
 import { SendingStatus, UserMessage } from '@sendbird/chat/message';
 import { isSameDay } from 'date-fns/isSameDay';
-import { useEffect, useMemo, useRef, useState } from 'react';
+import { useEffect, useMemo, useRef } from 'react';
 import styled from 'styled-components';
 
 import GroupChannelUI from '@uikit/modules/GroupChannel/components/GroupChannelUI';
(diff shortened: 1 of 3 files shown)


bang9:
Is it okay if I handle it in #312 ? There will be conflicts related to this, so I'll make the changes together

bang9:
https://github.com/sendbird/chat-ai-widget/pull/312/commits/7a26bee4e0fdf0a464c4c1ee86220d47a5b6cf53

When the message status is pending, the typing bubble doesn't appear because the message hasn't reached the server yet, even if it's just for a brief moment.

Since there is a subtle difference, I won't apply it immediately. It would be better to test whether the user experience is natural after implementing the display of the pending status, and then apply it if it works well.

chohongm:
> Is it okay if I handle it in #312 ? There will be conflicts related to this, so I'll make the changes together

yes!

chohongm:
> [7a26bee](https://github.com/sendbird/chat-ai-widget/commit/7a26bee4e0fdf0a464c4c1ee86220d47a5b6cf53)
> 
> When the message status is pending, the typing bubble doesn't appear because the message hasn't reached the server yet, even if it's just for a brief moment.
> 
> Since there is a subtle difference, I won't apply it immediately. It would be better to test whether the user experience is natural after implementing the display of the pending status, and then apply it if it works well.

We still needs this for auto-resend.

bang9:
> > [7a26bee](https://github.com/sendbird/chat-ai-widget/commit/7a26bee4e0fdf0a464c4c1ee86220d47a5b6cf53)
> > When the message status is pending, the typing bubble doesn't appear because the message hasn't reached the server yet, even if it's just for a brief moment.
> > Since there is a subtle difference, I won't apply it immediately. It would be better to test whether the user experience is natural after implementing the display of the pending status, and then apply it if it works well.
> 
> We still needs this for auto-resend.

That is block logic, not related to this(typing bubble).

PR #337, opened 2024-08-14
fix: do not disconnect in full screen mode

bang9:
## Changes
- do not trigger `useWidgetButtonActivityTimeout` in full screen mode


chohongm (approved):
LGTM

PR #356, opened 2024-09-10
[AC-3752] feat: add function call to carousel adapter interface

bang9:
## Changes
- Added adapter interface for function call data to carousel

ticket: [AC-3752]

## Additional Notes
- 

## Checklist
Before requesting a code review, please check the following:
- [x] **[Required]** CI has passed all checks.
- [x] **[Required]** A self-review has been conducted to ensure there are no minor mistakes.
- [x] **[Required]** Unnecessary comments/debugging code have been removed.
- [x] **[Required]** All requirements specified in the ticket have been accurately implemented.
- [x] Ensure the ticket has been updated with the sprint, status, and story points.

[AC-3752]: https://sendbird.atlassian.net/browse/AC-3752?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -81,12 +81,12 @@ export const DEFAULT_CONSTANT = {
   },
 } satisfies Partial<Constant>;
 
+// TODO: Remove this function when the Demo is finished
 function isMealsResponse(response: unknown): response is { meals: { strMeal: string; strMealThumb: string }[] } {
   return !!response && typeof response === 'object' && 'meals' in response && Array.isArray(response.meals);
 }
 
 type ConfigureSession = (sdk: SendbirdChat | SendbirdGroupChat | SendbirdOpenChat) => SessionHandler;
-
 type MatchString = string;
 type ReplaceString = string;
 
--- a/src/const.ts
+++ b/src/const.ts
@@ -10,7 +10,7 @@ import { StringSet } from '@uikit/ui/Label/stringSet';
 import type { ToggleButtonProps } from './components/widget/WidgetToggleButton';
 import { BotStyle } from './context/WidgetSettingContext';
 import RefreshIcon from './icons/ic-refresh.svg';
-import { SendbirdChatAICallbacks } from './types';
+import { FunctionCallAdapter, SendbirdChatAICallbacks, WidgetCarouselItem } from './types';
 import { noop } from './utils';
 
 // Most of browsers use a 32-bit signed integer as the maximum value for z-index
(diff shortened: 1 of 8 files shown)


src/types.ts
```diff
@@ -7,10 +7,8 @@ export interface SendbirdChatAICallbacks {
   onWidgetSettingFailure?: (error: Error) => void;
 }
 
-export interface FunctionCallRequestInfo {
-  headers: {
-    'Api-Token': string;
-  };
+export interface FunctionCallRequest {
+  headers: object;
```
chohongm, line 11: is there any particular reason why type changed to object?

src/const.ts
```diff
@@ -68,18 +68,25 @@ export const DEFAULT_CONSTANT = {
...
+
+        return [];
+      },
+    },
+  },
 } satisfies Partial<Constant>;
 
-type ConfigureSession = (sdk: SendbirdChat | SendbirdGroupChat | SendbirdOpenChat) => SessionHandler;
+function isMealsResponse(response: unknown): response is { meals: { strMeal: string; strMealThumb: string }[] } {
+  return !!response && typeof response === 'object' && 'meals' in response && Array.isArray(response.meals);
```
chohongm, line 86: !!response <- this is redundant no? typeof response === 'object' should cover it

src/const.ts
```diff
@@ -68,18 +68,25 @@ export const DEFAULT_CONSTANT = {
   messageInputControls: {
     blockWhileBotResponding: 10000,
   },
+  tools: {
+    functionCall: {
+      carouselAdapter({ response }) {
+        if (isMealsResponse(response)) {
+          return response.meals.map((it) => ({ title: it.strMeal, url: '', featured_image: it.strMealThumb }));
```
chohongm, line 75: can you explain either in ticket or in PR what this meals is?

src/const.ts
bang9: `null` is an object
```js
// true
typeof null === 'object';
```
so we should use `!!response` to filter it out

src/types.ts
bang9: The header varies depending on the function call set by the customer, just like other payloads.

src/const.ts
bang9: I'll share the context via DM

chohongm: This logic makes carouselAdapter is only for Noom usecase. Can it be used in other cases in the future? If so, we should make carouselAdapter more generic somehow.

chohongm: nvm I think its fine.

chohongm (approved):
LGTM

src/const.ts
bang9: Yes, that is an initial value. It can be overridden and will be removed.

PR #370, opened 2024-10-07
[SBTW-1920] Handling the issue of the parent window scrolling up when operating within an iframe

hello-mansoo:
## Changes
When using the Gen AI Interactive Demo on the Sendbird Official Site, there was an issue where the parent page would scroll up slightly when a new message arrived.

To address this, I disabled the logic in the Chat AI Widget that moves the page up by 150px when a new message arrives.

In local testing with various cases with iframes, I checked that the parent window does not move and the scrolling occurs only within the chat window when a new message arrives.

ticket: [https://sendbird.atlassian.net/browse/SBTW-1920]

## Additional Notes
- 

## Checklist
Before requesting a code review, please check the following:
- [ ] **[Required]** CI has passed all checks.
- [ ] **[Required]** A self-review has been conducted to ensure there are no minor mistakes.
- [ ] **[Required]** Unnecessary comments/debugging code have been removed.
- [ ] **[Required]** All requirements specified in the ticket have been accurately implemented.
- [ ] Ensure the ticket has been updated with the sprint, status, and story points.


src/components/CustomChannelComponent.tsx
```diff
@@ -241,13 +241,13 @@ export function CustomChannelComponent(props: CustomChannelComponentProps) {
...
-    bottomBuffer:
-      dynamicReplyOptions.length > 0 || isStaticReplyVisible ? 150 : 0,
-  });
+  // useScrollOnStreaming({
+  //   isLastBotMessage,
+  //   lastMessageRef,
+  //   // the reply panel component height is about 50px * 3(max replies) = 150px
+  //   bottomBuffer:
+  //     dynamicReplyOptions.length > 0 || isStaticReplyVisible ? 150 : 0,
+  // });
```
bang9: This hook ensures that the scroll moves to the bottom when messages are streamed. Would it be okay to remove this? I'm also curious if changing the bottomBuffer to `0` might resolve the issue instead

hello-mansoo: @bang9 
Oh, initially it behaved the same even when the bottomBuffer was set to 0, and it also scrolled correctly to the last message when the hook was commented out.

Upon reviewing the hook, I found that it works correctly when the block option of the `scrollIntoView` method is set to 'start' instead of 'end'.

It seems that the settings like 150 and the `scrollUtil()` method at the bottom were used to match the 'start' option.

I am attaching the Before and After videos.

There was an issue where the parent screen slightly scrolled when the first message was received, but now only the chat window inside the iframe scrolls, and it correctly follows when additional messages are received.

Before: https://drive.google.com/file/d/16vOe1LbAB63fxCTydKXNa4BDeCJhh_CV/view?usp=drive_link
After: https://drive.google.com/file/d/1bOTQRi66iHBWZ8A46VuSlqmguYRLzzJQ/view?usp=drive_link

Thanks for your detailed review, Airen:D

bang9 (approved):

bang9 (approved):

